PR #23, opened 2026-04-05
feat(ui): show current line number in status bar

rashpile:
## Summary

Adds a `L:42/380` indicator to the status bar showing the current cursor's line number and total lines in the file.

Closes umputun/revdiff#22

## What's changed

- New `lineNumberSegment()` displays current line position in the status bar
- Uses new file line number for context/add lines, old file line number for removed lines
- Positioned after hunk info: `filename | +X/-Y | hunk 2/5 | L:42/380`
- Degrades gracefully on narrow terminals — drops before hunk info at minimal level
- Hidden when focus is on tree pane or cursor is on a divider line

## Test plan

- [x] Unit tests for `lineNumberSegment` — context, add, remove, divider, empty, tree focus
- [x] Integration test verifying segment appears in status bar output
- [x] Degradation test confirming line number drops at minimal level
- [x] Full test suite passes

## Credits

**Powered by**

- Claude Code
- Ralphex

### DIFF
--- a/ui/model.go
+++ b/ui/model.go
@@ -914,7 +914,9 @@ func (m Model) hunkSegment() string {
 }
 
 // lineNumberSegment returns a formatted line number string like "L:42/380" for the status line.
-// returns empty string when focus is not on diff pane, cursor is out of range, or on a divider line.
+// The denominator is dynamic: on removed lines it shows the old file's max line number,
+// on context/added lines it shows the new file's max line number.
+// Returns empty string when focus is not on diff pane, cursor is out of range, or on a divider line.
 func (m Model) lineNumberSegment() string {
 	if m.focus != paneDiff {
 		return ""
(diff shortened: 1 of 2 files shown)


rashpile:
> 2. the denominator in `L:42/380` is the max line number in the diff, not total lines in file. for a partial diff this could be confusing (e.g., file has 500 lines but last hunk ends at 380). I'd drop the denominator and just show `L:42`, or rename to `L:42 of ~380` to signal it's approximate. what do you think?

I investigated and found the issue is actually slightly different than expected.

revdiff uses -U1000000 (full file context), so the denominator always covers the entire file, not just hunk boundaries. But the current code takes max(OldNum, NewNum) across all lines,  which means when lines are deleted from the end of a file, the denominator shows the old file length. For example, deleting 7 lines from the end of a 192-line file shows L:187/192 -  implying you can scroll further, but you can't.

  My proposed fix: use a dynamic denominator based on which side the cursor is on.
  - On context or added (+) lines → denominator = max NewNum (new file length)
  - On removed (-) lines → denominator = max OldNum (old file length)

  This way the denominator always matches the total line count of the file side you're currently looking at.

umputun:
good catch on the `-U1000000` context. the dynamic denominator based on change type makes sense, go ahead with that approach. also pls fix the 3 testifylint issues (`assert.Empty` instead of `assert.Equal(t, "", ...)`)

PR #41, opened 2026-04-06
feat: opencode config docs

hackmajoris:
# revdiff OpenCode integration

## Prerequisites

- [revdiff](https://github.com/umputun/revdiff) installed and in `PATH`
- One of the supported terminals: Ghostty, tmux, Kitty, WezTerm, cmux, iTerm2, Emacs vterm
- Clone the repo with plugin configs from https://github.com/umputun/revdiff
## Files

```
~/.config/opencode/
├── commands/
│   └── revdiff.md
├── tools/
│   ├── revdiff.ts
│   └── launch-revdiff.sh
└── plugins/
    ├── revdiff-plan-review.ts
    └── launch-plan-review.sh
```

## Installation

```sh
bash setup.sh
```

The script creates the target directories if needed, copies all files, marks the shell scripts as executable, and registers the plan-review plugin in `~/.config/opencode/opencode.json`. Or manually:

```sh
mkdir -p ~/.config/opencode/commands ~/.config/opencode/tools ~/.config/opencode/plugins
cp ../../.claude-plugin/skills/revdiff/scripts/launch-revdiff.sh ~/.config/opencode/tools/
chmod +x ~/.config/opencode/tools/launch-revdiff.sh
cp ../revdiff-planning/scripts/launch-plan-review.sh ~/.config/opencode/plugins/
chmod +x ~/.config/opencode/plugins/launch-plan-review.sh
cp commands/revdiff.md ~/.config/opencode/commands/
cp tools/revdiff.ts ~/.config/opencode/tools/
cp plugins/revdiff-plan-review.ts ~/.config/opencode/plugins/
```

Then register the plan-review plugin in `~/.config/opencode/opencode.json`:

```json
{
  "plugin": ["./plugins/revdiff-plan-review.ts"]
}
```

Restart OpenCode after installing — tools and commands are loaded at startup.


umputun (changes requested):
thx for the OpenCode integration, nice to see revdiff working with other tools.

the main issue is that `revdiff.ts` reimplements all the terminal detection logic (tmux, kitty, wezterm, cmux, ghostty, iTerm2, emacs vterm) that already lives in `launch-revdiff.sh`. every time we fix smth in the bash launcher (like the tmux `-T` compat fix we just merged), the TS version would be out of sync.

I'd suggest making the tool a thin wrapper around the existing launcher script instead. smth like:

```ts
const launcher = await $`which launch-revdiff.sh || echo ""`.text();
// or ship the script alongside the tool and call it directly
const result = await $`${launcher} ${ref}`.text();
```

the launcher already handles all terminals, sentinel polling, cleanup, and annotation capture. the TS tool just needs to call it and return stdout.

also a few bugs in the current TS if you keep it as-is:
1. Ghostty block doesn't set working directory, revdiff would open in the wrong dir
2. iTerm2 uses `current session` instead of targeting by `$ITERM_SESSION_ID` UUID, may hit the wrong pane
3. emacs vterm uses `term` instead of `vterm-mode`, won't work

umputun (commented):
much better now with the tool wrapping `launch-revdiff.sh` instead of reimplementing terminal detection, thx.

couple things I noticed:

1. `revdiff-plan-review.ts:111` - `if (!annotations!) return;` has a redundant non-null assertion after the negation. same on line 115 with `annotations!`. should be just `if (!annotations) return;` and `annotations` without `!`

2. `README.md:127` - `See 'plugins/opencode/README.md'` should be a proper markdown link

3. `tools/revdiff.ts` uses `Bun.$` directly while `revdiff-plan-review.ts` imports `{ $ } from "bun"`. minor inconsistency but worth picking one style

umputun (approved):
lgtm, thx for addressing the previous feedback. the tool wrapping `launch-revdiff.sh` is the right approach.

couple minor things to clean up when you get a chance:

1. `setup.sh` - `realpath --relative-to` isn't available on macOS default coreutils. the `|| echo` fallback works but produces stderr noise on every copy. smth like `${src#$REPO_ROOT/}` would be cleaner

2. `setup.sh` - when `opencode.json` doesn't exist, it just warns. could create a minimal `{"plugin": ["./plugins/revdiff-plan-review.ts"]}` so setup is fully automated

3. `revdiff-plan-review.ts` - several silent `catch {}` blocks (`isInstalled`, `launchReview`, `getLastPlanContent`, `injectAnnotations`). at least a `console.error` would help debugging when smth breaks

PR #43, opened 2026-04-06
Add visual range selection (V) and hunk annotation (H)

melonamin:
## Summary

- **Visual range selection** (`V`) — select a span of diff lines, confirm with Enter to annotate the range
- **Hunk annotation** (`H`) — one-key annotation of the entire hunk under the cursor

## Demo

https://github.com/user-attachments/assets/4b979ab7-9cd4-4e16-8a9a-cdf2965e7456

### DIFF
--- a/.claude-plugin/skills/revdiff/references/usage.md
+++ b/.claude-plugin/skills/revdiff/references/usage.md
@@ -87,9 +87,11 @@ When `--only` specifies a file that has no git changes (or when no git repo exis
 |-----|--------|
 | `a` or `Enter` (diff pane) | Annotate current diff line |
 | `A` | Add file-level annotation (stored at top of diff) |
+| `V` | Start visual range selection (navigate with `j`/`k`, then `a`/`Enter` to annotate) |
+| `H` | Annotate entire hunk under cursor |
 | `@` | Toggle annotation list popup (navigate and jump to any annotation) |
 | `d` | Delete annotation under cursor |
-| `Esc` | Cancel annotation input |
+| `Esc` | Cancel annotation input / cancel selection |
 
 **View:**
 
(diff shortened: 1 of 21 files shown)


umputun (commented):
the implementation is solid, good tests, clean data model. but I have a fundamental concern about whether this is worth the complexity.

**do we need range/hunk annotations?** the current single-line annotation already gives enough context. when someone writes `## file.go:43 (+) refactor this block`, the AI reading the annotations can see the surrounding code and infer the scope. I don't see a real use case where pointing at a specific line isn't sufficient. this adds ~2000 lines (including tests) for what seems like a nice-to-have.

as a user I love how the visual selection looks, it's a nice UX. but as a maintainer, 2k lines of new code to maintain for a feature without a clear problem statement is a concern.

I'd like to understand what problem you ran into that line annotations couldn't solve. if there's a concrete use case I'm missing, happy to reconsider.

if we do proceed, a few things to fix:

1. `theme/bundled.go` has 5 stray hex values without key names (bare `#44475a` etc.). they're parsed as comments and silently ignored, so `SelectedBg` is never set in any bundled theme. these need to be proper `color-selected-bg = #hex` entries, or removed if the selection background approach was dropped in favor of gutter indicators
2. whitespace-only alignment changes in `cmd/revdiff/main.go`, `theme/theme.go`, `ui/styles.go` that aren't related to the feature. pls revert those
3. `A` (file-level annotation) should be blocked during selection mode, like point annotation is

umputun:
closing this one. the "hunk" keyword approach from #47 ended up covering the main use case with ~50 lines instead of a full visual selection mode. thx for the PR and the discussion though, it helped shape the simpler solution.

PR #63, opened 2026-04-08
Add vim-style viewport scroll (zz/zt/zb) and quit aliases (ZZ/ZQ)

anpryl:
## Summary
- Add multi-key sequence support to the keymap system for two-key vim bindings
- New viewport scroll bindings: `zz` (center), `zt` (top), `zb` (bottom) — work in both diff and tree panes
- New quit aliases: `ZZ` (save & quit), `ZQ` (discard & quit)
- 3 new remappable actions: `scroll_center`, `scroll_top`, `scroll_bottom`
- `HasPrefix()` method on `Keymap` enables prefix key detection without hardcoding

## Changes
- `app/keymap/keymap.go` — new actions, default bindings, `HasPrefix()` method
- `app/keymap/keymap_test.go` — tests for new actions, prefix detection, multi-key bindings
- `app/ui/model.go` — `pendingKey` field for multi-key buffering in `handleKey`, scroll actions in diff/tree nav
- `app/ui/diffview.go` — `bottomAlignViewportOnCursor()` (center and top already existed)
- `README.md`, `site/docs.html`, plugin reference docs — updated keybinding tables

## Test plan
- [x] All existing keymap tests pass
- [x] All existing UI tests pass
- [x] New `TestHasPrefix` and `TestMultiKeyBindings` tests pass
- [x] Manual: press `zz`/`zt`/`zb` in diff pane — viewport repositions without moving cursor
- [x] Manual: press `zz`/`zt`/`zb` in tree pane — tree scroll repositions without moving cursor
- [x] Manual: press `ZZ` — quits (same as `q`)
- [x] Manual: press `ZQ` — discards and quits (same as `Q`)
- [x] Manual: press `z` then an unmapped key (e.g., `zx`) — prefix discarded, no action
- [x] Manual: `--dump-keys` shows zz/zt/zb/ZZ/ZQ bindings

🤖 Generated with [Claude Code](https://claude.com/claude-code)


umputun (commented):
thx for the PR, both the vim viewport scroll and the multi-key sequence support are welcome additions.

however, I think this should be split into two separate PRs because it mixes two independent concerns:

**PR 1 - viewport scroll actions** (`scroll_center`, `scroll_top`, `scroll_bottom`, sidepane methods, docs). these are pure view actions that can work with single-key bindings today. `centerViewportOnCursor` and `topAlignViewportOnCursor` already exist on master, so only `bottomAlign` is new. bind them to single keys for now (whatever makes sense), add the sidepane scroll methods, update docs - done.

**PR 2 - multi-key sequence infrastructure** (`pendingKey` buffering, `HasPrefix`, handler signature change `KeyMsg` → `Action`, `ZZ`/`ZQ` aliases). this is a general keyboard handling extension, not specific to viewport scroll. once merged, the viewport scroll actions from PR 1 can be rebound to `zz`/`zt`/`zb` as defaults, and `ZZ`/`ZQ` quit aliases come along for free.

the reason I want them split: these are two distinct, technically unrelated changes, each self-contained and much easier to review separately. viewport scroll actions don't need multi-key infrastructure (they work with single-key bindings), and multi-key infrastructure doesn't need viewport scroll to be useful. beyond that, the multi-key infrastructure has its own design questions that shouldn't block the viewport scroll feature, and the implementation is partial - it handles printable character sequences (`zz`, `ZQ`) but not leader-key combos (e.g., `ctrl+b a`). that's fine for a first step, but it's worth discussing scope and direction separately.

**concerns about the multi-key infrastructure (PR 2):**

1. `pendingKey` has no timeout - if user presses `z` and walks away, the stale prefix persists indefinitely until next keypress eats it silently. vim itself has `timeoutlen` for this
2. no feedback when prefix is buffered - user presses `z` and nothing visible happens, no status bar indicator
3. `HasPrefix` does linear scan with no conflict detection - a future binding to `z` as a single key would silently break `zz`/`zt`/`zb`
4. non-Latin keyboard layouts won't work with multi-key sequences (Copilot caught this too) - `layoutResolve` only handles single runes, so `pendingKey` concatenation produces untranslated strings
5. the handler signature change from `tea.KeyMsg` to `keymap.Action` is a one-way door - any future handler needing the raw key (text input, modifier detection) would need both params threaded through

none of these are blockers per se, but they deserve their own focused discussion without holding up the viewport scroll.

**concerns about viewport scroll (PR 1):**

1. `ScrollToCenter` and `ScrollToBottom` in both `FileTree` and `TOC` don't clamp offset to `max(0, len(entries)-height)` - when cursor is near the end, offset goes past valid range leaving blank space at the bottom. one-liner fix each
2. no unit tests for the new sidepane scroll methods - the offset clamping bug above shows why these are needed
3. CI build is failing - `gocyclo` reports complexity 21 (threshold 20) on `handleKey`, and `revive` flags an empty else block at `model.go:576`. these are from the multi-key logic, so splitting the PR would resolve them naturally

PR #73, opened 2026-04-08
feat: intra-line word-diff highlighting (experimental)

rashpile:
## Summary

Adds optional GitHub-style intra-line highlighting for paired remove/add lines. Within a hunk, paired `-`/`+` lines run through a token-level diff and only the *actually changed* spans get a brighter background overlay — the existing whole-line add/remove styling is preserved.

Off by default. Toggle with `W` or start with `--word-diff`.

## Why

When a line is edited rather than wholly added/removed, the user has to eyeball which tokens changed. This is especially painful in dense refactors and indentation-only changes. Word-level highlights make the actual edit pop.

## What's in

- New `app/ui/worddiff.go` — pairing, range computation, ANSI insertion helper, and `highlightIntraLineChanges` overlay
- Pairing logic extracted from `buildModifiedSet` into `pairHunkLines` (`app/ui/collapsed.go`); equal runs zip 1:1, uneven runs match by `2*commonPrefix + 2*commonSuffix` similarity
- Token-level diff via `github.com/sergi/go-diff/diffmatchpatch` with `DiffCleanupSemantic`
- 30% similarity threshold: pairs whose common spans cover < 30% of the shorter line are treated as full replacements and get **no** intra-line overlay (avoids noisy spurious matches like "ind", "diff" inside otherwise unrelated lines)
- Two new theme color keys: `color-word-add-bg`, `color-word-remove-bg` (defaults `#4a7a1a` / `#a03838`); wired into all 5 bundled themes
- New `--word-diff` CLI flag and `W` keybinding (`ActionToggleWordDiff`)
- Status bar mode icon (⇄) when active
- `--no-colors` fallback uses reverse-video markers, matching the search-highlight fallback
- Works alongside wrap, search, blame, line-numbers, collapsed mode (intra-line ANSI is inserted before `extendLineBg` and survives `ansi.Wrap` automatically)
- Unit tests for pairing, range computation, ANSI insertion, and rendering integration; coverage for the new package is ~94%

## Open questions for review

I can see this helping reviewers quickly spot what changed inside a line, but I'd land it as **experimental** because:

  - Needs validation on a large corpus of real diffs before trusting the output
  - The pairing heuristic and similarity threshold need real-world tuning
  - Color themes might needed tuning
  - Computational overhead per file load (small but non-zero — runs `diffmatchpatch.DiffMain` on every paired line)
  - **Is this worth shipping at all?** Should it be on by default, opt-in via flag, or dropped? I don't want to add a feature plus an ongoing tuning loop and a similarity threshold to maintain if in practice nobody uses it.


umputun:
thx for this, and especially for the detailed comparison analysis you posted — it made the path forward obvious.

merged a combined version in #87 that takes the key ideas from this PR:

- `pairHunkLines` with prefix/suffix scoring, shared between word-diff and collapsed mode's `buildModifiedSet`
- 30% similarity gate (non-whitespace tokens) to suppress noise on force-paired dissimilar lines
- dedicated `color-word-add-bg` / `color-word-remove-bg` theme keys with HSL auto-derivation via `shiftLightness`
- reuse of `insertHighlightMarkers` for both search and word-diff overlays

paired with @daulet's zero-dependency token-level algorithm from #74 (regex tokenizer + LCS) so we avoided the `go-diff` vendored dep.

the final version is always-on (no toggle flag, no W keybinding, no status icon) to keep the surface minimal. closing this one as its ideas shipped in #87.

PR #74, opened 2026-04-08
feat: add intra line highlights

daulet:
When left and right lines are similar enough, highlight what changed. 

Before:
image

After:
image

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,6 +8,7 @@ Built for a specific use case: reviewing code changes, plans, and documents with
 
 - Structured annotation output to stdout - pipe into AI agents, scripts, or other tools
 - Full-file diff view with syntax highlighting
+- Intraline change emphasis on related remove/add lines (delta-style within-line hints in unified view)
 - Collapsed diff mode: shows final text with change markers, toggle with `v`
 - Word wrap mode: wraps long lines at viewport boundary with `↪` continuation markers, toggle with `w`
 - Line numbers: side-by-side old/new line number gutter, toggle with `L`
--- a/app/ui/diffview.go
+++ b/app/ui/diffview.go
@@ -290,6 +290,10 @@ func (m Model) prepareLineContent(idx int, dl diff.DiffLine) (lineContent, textC
 	if hasHighlight {
 		textContent = strings.ReplaceAll(m.highlightedLines[idx], "\t", m.tabSpaces)
 	}
+	// only inject intraline markers when highlighted output still maps 1:1 to the raw text.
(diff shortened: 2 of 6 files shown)


umputun (commented):
thx for the PR. heads up - #73 by @rashpile is also implementing intra-line highlights. some key differences between the two:

- #73 (+3366/-169, 45 files) has opt-in toggle (`W` key + `--word-diff` flag), dedicated theme color keys (`WordAddBg`/`WordRemoveBg`), status bar icon, collapsed mode integration, ~94% test coverage, full docs update. uses `sergi/go-diff` as a new dependency
- this PR (+374/-10, 6 files) is much smaller footprint, zero new deps (hand-rolled LCS + Levenshtein), cleaner separation in a single file. always-on by design

a few things to address in this approach:

1. **reuses `SearchBg` for emphasis** - when search is active on a highlighted line, intra-line markers and search highlights become indistinguishable. needs separate colors
2. **naive mode threshold 1.0** - when remove/add counts are equal, lines are always paired regardless of similarity. two completely unrelated lines get noisy intra-line highlights
3. **CI failing** - exhaustive switch lint error in `intraline.go:48`, missing `ChangeContext`/`ChangeDivider` cases

the small footprint and zero-dependency approach is a real advantage here. the hand-rolled LCS + Levenshtein keeps things self-contained, which I value.

would be great if you and @rashpile could sync up on this - two parallel implementations of the same feature. there might be synergy, e.g. combining the smaller footprint from here with some of the integration work from #73, or joining forces on one PR.

umputun:
thx for this. merged a combined version in #87 that takes the core algorithm from this PR:

- regex tokenizer (`[\pL\pN_]+|\s+|[^\pL\pN_\s]+`) splitting into word/whitespace/punctuation tokens
- LCS DP + backtrace for `lcsKeptTokens`
- byte-offset based `matchRange` ranges, reusing the existing infrastructure
- zero new dependencies

paired with @rashpile's production integration from #73 (pairing heuristic, 30% similarity gate, dedicated theme colors with HSL auto-derivation, collapsed mode refactor).

few changes on top of your implementation:
- replaced the `0.60 / 1.00` Levenshtein pairing with prefix/suffix scoring (avoids forced pairing on same-count blocks with unrelated content)
- added a 500-byte cap in `changedTokenRanges` to prevent O(m*n) LCS memory blowup on minified content
- dedicated word-diff bg colors instead of reusing `SearchBg` (search + word-diff can now coexist without visual collision)

the final version is always-on matching your design. closing this one as its core algorithm shipped in #87.

PR #77, opened 2026-04-09
feat(zed): add task for running revdiff

rashpile:
**Summary**

  - Add "revdiff current file" Zed task that opens the focused file in revdiff TUI
  - Add Zed tasks for branch diff and uncommitted review
  - Annotations are written to a temp file via --output and copied to clipboard on quit
  - Cleans up stale output before each run to avoid copying old annotations
  - Includes keybinding instructions for cmd+shift+d

**Zed task config**

`  "command": "rm -f /tmp/revdiff-out.txt; revdiff --only=$ZED_FILE --output=/tmp/revdiff-out.txt; [ -f /tmp/revdiff-out.txt ] && pbcopy < /tmp/revdiff-out.txt; true"`

  Keybinding (user-side, ~/.config/zed/keymap.json)

  `{ "context": "Workspace", "bindings": { "cmd-shift-d": ["task::Spawn", { "task_name": "revdiff current file" }] } }`


umputun (commented):
the tasks themselves look good, but `.zed/tasks.json` in the repo is for revdiff developers (lint, test, etc.), not the place where users would discover Zed integration.

I'd suggest moving this into documentation instead, e.g. a "Zed" section in `site/docs.html` or `README.md` alongside other editor integrations. that way users actually find it, and the task config becomes a copy-paste snippet they add to their own project or user-level Zed config.

couple things to fix in the snippets too:

1. `pbcopy` is macOS-only, needs to either be made cross-platform (e.g. `pbcopy 2>/dev/null || xclip -selection clipboard 2>/dev/null || xsel --clipboard`) or at minimum documented as macOS-only
2. `$ZED_FILE` should be quoted (`"$ZED_FILE"`) to handle filenames with spaces

would you be up for converting this into a docs addition?

umputun (commented):
looks good, moved to docs as discussed. two small things still:

1. `$ZED_FILE` needs quoting (`"$ZED_FILE"`) in all three task commands - paths with spaces will break otherwise
2. the three JSON task objects are shown as standalone blocks without array wrapping. `.zed/tasks.json` expects a JSON array, so either wrap them in `[...]` with commas, or add a note like "add each object to the tasks array"

PR #90, opened 2026-04-10
Add Mercurial VCS support

paskal:
## Summary

- Auto-detect Mercurial repos via `DetectVCS()` walking up directories for `.hg`
- `Hg` struct implements `Renderer` and `Blamer` interfaces using `hg status`, `hg diff --git`, and `hg annotate`
- Git-style ref translation: `HEAD`→`.`, `HEAD~N`→`.~N`, `HEAD^`→`.^`, `HEAD^N`→`pN(.)`
- `--staged` warns when used with hg (no staging area), `--all-files` returns error (git only)
- Shared `runVCS()` helper eliminates duplicate command execution between Git and Hg
- Documentation updated across README, CLAUDE.md, site/docs.html, site/index.html, plugin references
- 37 new test cases covering VCS detection, renderer, blame, ref translation, and E2E with real hg repos

### DIFF
--- a/.claude-plugin/skills/revdiff/references/config.md
+++ b/.claude-plugin/skills/revdiff/references/config.md
@@ -26,7 +26,7 @@ Then uncomment and edit the values you want to change.
 | `--wrap` | `REVDIFF_WRAP` | Enable line wrapping in diff view | `false` |
 | `--collapsed` | `REVDIFF_COLLAPSED` | Start in collapsed diff mode | `false` |
 | `--line-numbers` | `REVDIFF_LINE_NUMBERS` | Show line numbers in diff gutter | `false` |
-| `--blame` | `REVDIFF_BLAME` | Show git blame gutter on startup | `false` |
+| `--blame` | `REVDIFF_BLAME` | Show blame gutter on startup | `false` |
 | `--word-diff` | `REVDIFF_WORD_DIFF` | Highlight intra-line word-level changes in paired add/remove lines | `false` |
 | `--no-confirm-discard` | `REVDIFF_NO_CONFIRM_DISCARD` | Skip confirmation when discarding annotations with Q | `false` |
 | `--chroma-style` | `REVDIFF_CHROMA_STYLE` | Chroma color theme for syntax highlighting | `catppuccin-macchiato` |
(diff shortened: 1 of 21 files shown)


umputun (changes requested):
Thanks for the PR. Overall the approach is solid — reusing `Renderer`/`Blamer` interfaces, the `runVCS` extraction, and the E2E tests with real `hg` repos are all good choices. Tests pass, lint clean, race clean.

A few things to address before this can merge:

**1. `--only` in hg repos bypasses hg diff entirely** (`app/main.go:565`)

The git path uses `FallbackRenderer` which delegates to the git renderer first and only falls back to disk for files not in the diff. The hg path switches to `FileReader` unconditionally, so `revdiff --only foo.txt HEAD~1` in an hg repo shows the current file contents instead of the actual hg diff. This is a meaningful behavior asymmetry that breaks a common workflow.

Either add an `HgFallbackRenderer` analogous to `FallbackRenderer`, or pass the `*Hg` instance and let the existing model-side `--only` filtering handle path matching. The latter is simpler if it covers the needed cases.

**2. `revFlag` triple-dot bug** (`app/diff/hg.go:109`)

`strings.Cut(ref, "..")` matches the first `..` in `"main...feature"`, producing `left="main"`, `right=".feature"`, which yields `hg status -r main -r .feature` — invalid revset. Interestingly, you got this right in `hgblame.go:35-41` where `blameTargetRef` checks `...` before `..`. The same guard is missing from `revFlag`.

Mirror the `blameTargetRef` pattern:

```go
func (h *Hg) revFlag(flag, ref string) []string {
	if ref == "" {
		return nil
	}
	// check triple-dot first, treat as ancestor..right in hg terms
	if left, right, ok := strings.Cut(ref, "..."); ok {
		// handle triple-dot (choose the semantics you want)
		...
	}
	if left, right, ok := strings.Cut(ref, ".."); ok {
		...
	}
	return []string{flag, translateRef(ref)}
}
```

Also add a test case in `TestHg_RevFlag` for the triple-dot input.

**3. Style: `parseStatus` and `hgStatusToFileStatus` should be methods** (`app/diff/hg.go:40, 85`)

Both are called exclusively from methods on `Hg` (`ChangedFiles`, `changedFilesFromDiff`, and `parseStatus` respectively). Per the project's CLAUDE.md:

> Functions called only from methods MUST be methods ... This rule applies regardless of whether the function needs access to struct fields - field access is irrelevant to this decision.

Convert them to `(h *Hg) parseStatus(...)` and `(h *Hg) hgStatusToFileStatus(...)`. Mechanical change.

**Note on Copilot finding #3** (`TestDetectVCS_Git`): I don't think this one needs a fix — the test has an explicit comment that it runs inside the revdiff repo, and hermetic unit tests already exist in `app/diff/vcs_test.go`. It's a deliberate integration smoke test.

Once 1-3 are addressed, this should be ready to merge.

paskal:
Please take another look

umputun (approved):
LGTM, thx

PR #91, opened 2026-04-11
Add lint-scripts CI job, fix SC2181 shellcheck warnings

paskal:
## Summary

- Add `lint-scripts` CI job: validates YAML frontmatter via standalone `.github/scripts/validate-frontmatter.py` and runs `shellcheck` on all `.sh` files (excludes `vendor/`)
- Fix SC2181 in 3 scripts: replace `if [ $? -ne 0 ]` after command substitution with `if !` idiom
  - `.claude-plugin/skills/revdiff/scripts/launch-revdiff.sh`
  - `plugins/codex/skills/revdiff/scripts/launch-revdiff.sh`
  - `plugins/revdiff-planning/scripts/launch-plan-review.sh`

Mirrors umputun/cc-thingz#10 CI checks for this repo. No shfmt reformat — only the SC2181 fix.

### DIFF
--- a/.claude-plugin/skills/revdiff/scripts/launch-revdiff.sh
+++ b/.claude-plugin/skills/revdiff/scripts/launch-revdiff.sh
@@ -193,7 +193,7 @@ $REVDIFF_CMD; touch $(sq "$SENTINEL")
 LAUNCHER
     chmod +x "$LAUNCH_SCRIPT"
 
-    GHOSTTY_TERM_ID=$(osascript - "$LAUNCH_SCRIPT" "$CWD" <<'APPLESCRIPT'
+    if ! GHOSTTY_TERM_ID=$(osascript - "$LAUNCH_SCRIPT" "$CWD" <<'APPLESCRIPT'
 on run argv
     set launchScript to item 1 of argv
     set cwd to item 2 of argv
@@ -209,8 +209,7 @@ on run argv
     end tell
 end run
 APPLESCRIPT
-    )
-    if [ $? -ne 0 ]; then
+    ); then
         rm -f "$SENTINEL" "$LAUNCH_SCRIPT"
         exit 1
     fi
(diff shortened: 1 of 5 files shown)


umputun (commented):
thx for porting this from cc-thingz. two things before I can merge:

**1. the reformat doesn't belong here.** the PR title says "fix SC2181" but 95% of the diff in the three scripts is a full `shfmt` reformat (4-space → tabs, compact redirections, `case` rewrap). the actual SC2181 fix is only a handful of lines per file. pls drop the reformat and keep just the `if ! cmd` change. if you want to push tab-style + shfmt as a separate discussion, that's a different PR and we'd also need to commit a shfmt config so the style is enforced going forward, otherwise the next editor just undoes it.

**2. the checkout/setup-go reorder in the `build` job is unrelated scope.** it may be a legit fix but it has nothing to do with lint-scripts or SC2181. pls pull it out into its own PR if you want it in.

minor stuff on the new `lint-scripts` job (not blockers, but worth addressing while we're here):

- the inline python heredoc in `ci.yml` is 30 lines of code embedded in YAML. would prefer it as a standalone `.github/scripts/validate-frontmatter.py` so it can be tested locally and doesn't clutter the workflow
- the frontmatter walker doesn't prune `vendor/` — works today because no vendored `.md` has frontmatter, but `go mod vendor` could pull one in tomorrow. pls add vendor to the skip filter, ideally via `dirs[:] = [d for d in dirs if d not in ('.git', 'vendor')]` so the walk is pruned at source instead of filtered per-iteration
- `lint-scripts` checkout is missing `persist-credentials: false` — the `build` job has it, should be consistent

copilot's 3 inline comments on the first commit look already addressed in the follow-up.

I'd like to see just the SC2181 fix in isolation first, then we can talk about the reformat and the build-job reorder separately.

paskal:
Addressed the feedback, extra reindent came from shfmt.

umputun (approved):
lgtm, thx

PR #94, opened 2026-04-11
feat: add REVDIFF_LAUNCHER for custom terminal override

nborodikhin:
## Summary

Claude Code launches hook programs without a terminal. Its integrations currently use a number of custom approaches to launch TUI programs in different terminal emulators.

This works but there are some drawbacks:
- the user temporarily loses access to the Claude Code session and is not able to, say, scroll around to review conversation history
- the overlay is not a pop-up **window** in the window manager sense, and does not stand out in the application switcher (e.g., Alt-Tab)
- the user might want to use a different approach, for example create a separate terminal window and launch `revdiff` in it

## Changes

- Add `REVDIFF_LAUNCHER` env var to bypass all built-in terminal detection and delegate to a user-supplied launcher script
- The script receives the revdiff binary + args as `$@`; the window title is passed via `$REVDIFF_TITLE`
- Applied to both `launch-revdiff.sh` and `launch-plan-review.sh`
- Docs updated in README and Claude Code plugin references

## Usage

For example, to launch a separate Kitty window instead of an overlay:
```sh
# ~/bin/my-revdiff-launcher
#!/bin/sh
exec kitty -T "$REVDIFF_TITLE" "$@"
```

Start Claude Code with a custom revdiff launcher:
```bash
$ export REVDIFF_LAUNCHER="$HOME/bin/my-revdiff-launcher"
$ claude
```

## Test plan

- [x] Set `REVDIFF_LAUNCHER` to a custom script and verify it is invoked instead of built-in terminal detection
- [x] Verify annotations are still captured correctly via the custom launcher
- [x] Verify `REVDIFF_TITLE` is available in the launcher script
- [x] Verify unset `REVDIFF_LAUNCHER` falls through to normal terminal detection unchanged

### DIFF
--- a/.claude-plugin/skills/revdiff/references/config.md
+++ b/.claude-plugin/skills/revdiff/references/config.md
@@ -46,14 +46,23 @@ Then uncomment and edit the values you want to change.
 | `--config` | `REVDIFF_CONFIG` | Path to config file | `~/.config/revdiff/config` |
 | `--dump-config` | | Print default config to stdout and exit | |
 
-## Popup Size (Claude Code plugin)
+## Popup Size and Custom Launcher (Claude Code plugin)
 
-When launched via the Claude Code plugin skill, revdiff opens in a terminal overlay. The popup size is configurable via env vars:
+When launched via the Claude Code plugin skill, revdiff opens in a terminal overlay. The overlay behavior is configurable via env vars:
 
 | Env var | Description | Default |
 |---------|-------------|---------|
 | `REVDIFF_POPUP_WIDTH` | Tmux popup width (e.g., `100%`, `80%`) | `90%` |
 | `REVDIFF_POPUP_HEIGHT` | Tmux popup height / wezterm split percent | `90%` |
+| `REVDIFF_LAUNCHER` | Path to a custom launcher script (overrides all built-in terminal detection) | unset |
+
+When `REVDIFF_LAUNCHER` is set, the script is called with the revdiff binary and all args as `$@`. The window title is available as `$REVDIFF_TITLE`.
+
+Example `~/bin/my-revdiff-launcher`:
+```sh
+#!/bin/sh
+exec kitty -t "$REVDIFF_TITLE" "$@"
+```
 
 ## Themes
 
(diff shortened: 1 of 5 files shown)


umputun:
Thanks for the PR, but I'm going to close this one.

Two concrete problems.

First, per CONTRIBUTING.md, first-time PRs need an issue or discussion first so we can align on whether the idea fits the project before anyone writes code. That wasn't done here.

Second, **the documented example leaves a hung wrapper on macOS**. I tested the exact example from the PR body verbatim:

```sh
#!/bin/sh
exec kitty -T "$REVDIFF_TITLE" "$@"
```

What happens end-to-end through `launch-revdiff.sh`:

1. Launcher fires, `sh -c` runs it with the correct `REVDIFF_TITLE` and `$@ = revdiff --output=/tmp/… HEAD~5`
2. `exec kitty ...` spawns a new kitty process; a window opens; the review UI works; I annotate a line
3. I close the kitty window via the window manager
4. kitty.app stays alive in the background (standard macOS `.app` lifecycle — the dot stays in the dock until Cmd+Q)
5. The `kitty` CLI process the launcher exec'd is still a live child of that still-alive app
6. `sh -c` in `launch-revdiff.sh` blocks forever waiting for it
7. The wrapper never reaches `cat "$OUTPUT_FILE"`, never returns, never exits
8. Claude's bash tool call eventually hits its 10-minute timeout, leaving an orphan kitty.app + hung wrapper per invocation

revdiff itself does write the annotation to its output file atomically on exit, so the skill's Step 3 timeout-fallback in `SKILL.md` can eventually scrape annotations out of `$TMPDIR/revdiff-output-*`. But that fallback was designed for long reviews hitting the 10-minute cap, not for a launcher contract that permanently hangs. Using this example in practice = a ~10-minute wait per review plus an orphan `kitty.app` every single time.

Your test plan items all check properties that do pass (launcher fires, title propagates, annotations get captured, fall-through when unset). None of them verify the one property that actually matters for the plugin contract: **does the wrapper return control to the plugin after the user closes the review window?** That's how this slipped through.

To make the feature work cleanly you'd need a curated set of launch strategies rather than an open-ended `sh -c` hook — each one knowing how to open a terminal *and* block on the window's lifetime rather than the app's. That's not trivial, especially on macOS where separating window lifetime from app lifetime usually requires AppleScript or spawning a disposable helper app.

Please open a Discussion describing the concrete workflow you want to enable and the constraints you care about, and we can figure out a shape that works cross-platform before anyone writes more code.

nborodikhin:
A better approach is implemented by #126

PR #98, opened 2026-04-12
fix(ui): show single-column line numbers for full-context files

rashpile:
## Summary
- In full-context modes (`--only`, `--stdin`, `--all-files`), line number gutter now shows a single column (`" NNN"`) instead of two identical columns (`" NNN NNN"`), saving horizontal space
- Detection is per-file via `isFullContext()` in `handleFileLoaded`, so files with real diffs still get two-column gutters
- All rendering paths (expanded, wrapped, collapsed, wrapped-collapsed) go through shared `lineNumGutter()`/`lineNumGutterWidth()` — no special-casing needed

## Test plan
- [x] Unit tests for `singleColLineNum` detection (full-context vs diff files)
- [x] Unit tests for `lineNumGutterWidth()` in single-column and two-column modes
- [x] Unit tests for `lineNumGutter()` output format in both modes
- [x] Width-consistency test: gutter string width == `lineNumGutterWidth()` for all line types in both modes
- [x] Full test suite passes (87.8% coverage), linter clean
- [x] Manual: `revdiff --only README.md` + `L` toggle — single column
- [x] Manual: `revdiff` on branch with changes + `L` — two columns
- [x] Manual: `--only` on file that IS in the diff — two columns
- [x] Manual: blame (`B`) + line numbers (`L`) together in full-context mode


umputun (commented):
lgtm on the approach, clean implementation. Reusing `isFullContext()` for both TOC and single-col detection is the right call, and the width-consistency test is solid.

a couple of things before merge:

1. codex copy of `usage.md` wasn't synced. CLAUDE.md requires `.claude-plugin/skills/revdiff/references/usage.md` and `plugins/codex/skills/revdiff/references/usage.md` to stay byte-identical. The codex copy still has the old "side-by-side old/new numbers in gutter" line.

2. `app/ui/model.go` needs `gofmt`. The new `singleColLineNum` field added extra padding but the prior 3 lines (`showHelp`, `wrapMode`, `crossFileHunks`) weren't re-aligned. `gofmt -l` flags the file. Linter didn't catch it because `gofmt` isn't in the golangci-lint config, but `make fmt` or `gofmt -s -w` fixes it.

minor, non-blocking: `TestHandleFileLoaded_SingleColLineNum_FullContext` only asserts the flag is set, not that the rendered output is narrow. The width-consistency test covers the invariant though, so the integration gap is mostly theoretical. Feel free to skip.

rashpile:
Fixed all three items:

  1. **Codex usage.md synced** — line 111 updated to match claude-plugin copy, now byte-identical (`ffebf21`)
  2. **gofmt alignment** — `showHelp`/`wrapMode`/`crossFileHunks` re-aligned with the new field (`ffebf21`)
  3. **Integration test gap** — `TestHandleFileLoaded_SingleColLineNum_FullContext` now asserts `lineNumGutterWidth() == W+1` after file load (`11dd05e`)

umputun (approved):
lgtm, thx

PR #101, opened 2026-04-13
docs: add AUR install instructions

kovstas:
Packaged revdiff for the AUR for convenience - the PKGBUILD pulls prebuilt binaries from the GitHub Releases page.
https://aur.archlinux.org/packages/revdiff

This PR just adds a short note to the README so Arch users know they can install it via paru -S revdiff (or any other AUR helper).


umputun (approved):
lgtm, thx for packaging this for AUR.

I'll also add deb/rpm install instructions since we ship those in releases too.

PR #103, opened 2026-04-13
Add `--include` prefix filter flag

rashpile:
Add `--include` (`-I`) flag for prefix-based file inclusion, complementing the existing `--exclude` (`-X`). Both flags compose: include narrows the file set first, then exclude removes from it.

```
revdiff --include src                          # only src/ files
revdiff --include src --exclude src/vendor     # src/ minus src/vendor/
revdiff --all-files --include app              # all tracked files under app/
```

### Changes

**Core filter** (`app/diff/`)
- Extract shared `matchesPrefix` and `normalizePrefixes` helpers into `prefix.go`, refactor `ExcludeFilter` to use them
- Add `IncludeFilter` in `include.go` — same decorator pattern, inverted logic (keep matching files)

**CLI wiring** (`app/main.go`)
- `--include` / `-I` flag with config file (`ini-name`), env var (`REVDIFF_INCLUDE`), and comma-delimited support
- Mutual exclusivity: `--include` + `--only` and `--stdin` + `--include` are errors
- Composition order in all three renderer factories: inner -> IncludeFilter -> ExcludeFilter

**Tests** (`app/diff/include_test.go`, `app/main_test.go`)
- Include filter unit tests: prefix matching, normalization, ChangedFiles filtering, FileDiff passthrough, error propagation
- CLI parsing tests: long flag, short flag, env var, config file
- Validation tests: stdin + include, only + include
- Renderer factory tests: git/hg with include, all-files with include, include + exclude functional composition

**Docs**
- README, site/docs.html, plugin references (claude, codex, pi)


umputun (commented):
lgtm overall, clean implementation that mirrors `ExcludeFilter` well. tests pass, linter clean, no races.

couple things:

1. `prefix.go` - I'd prefer moving `matchesPrefix` and `normalizePrefixes` into `diff.go` instead of a separate file. two small functions shared by both filters don't justify a new file. tests for them should move to `diff_test.go` accordingly

2. duplicate tests - `TestMatchesPrefix` and `TestNormalizePrefixes` exist in both `include_test.go` and `exclude_test.go`. move them to `diff_test.go` (with the functions) and drop both duplicates

3. minor: `makeNoVCSRenderer` accepts `include` param but it's unreachable - no-VCS mode requires `--only`, and `--include` + `--only` is rejected by `parseArgs`. not a blocker, `--exclude` has the same pattern, but worth knowing

the `--include` / `--only` mutual exclusivity makes sense, they have different semantics (prefix filter vs exact/suffix file selection)

rashpile:
All issues addressed:

   - Moved `matchesPrefix` and `normalizePrefixes` from `prefix.go` into `diff.go`, deleted `prefix.go`
   - Moved shared tests (`TestMatchesPrefix`, `TestNormalizePrefixes`) into `diff_test.go`, dropped duplicates from `include_test.go` and `exclude_test.go`
   - Removed unreachable `include`/`exclude` params from `makeNoVCSRenderer` — no-VCS mode requires `--only`, which is mutually exclusive with `--include`, and `--exclude` on `FileReader` is a no-op

umputun (approved):
lgtm, all review points addressed. thx for the cleanup on `makeNoVCSRenderer` - removing both unreachable params is cleaner than what I suggested.

one non-blocking note from Copilot: if all include prefixes normalize to empty (e.g. `REVDIFF_INCLUDE=,`), `matchesPrefix` returns false for everything and all files silently disappear. unlikely edge case, but `--exclude` is harmless in the same scenario (nothing excluded = no-op) while `--include` is destructive (nothing included = empty list). something to keep in mind for later, not a blocker.

rashpile:
fix: handle empty normalized include prefixes as no-op

PR #108, opened 2026-04-15
fix mktemp file/path.XXXX.suffix bug

tsimoshka:
`mktemp`, at least on macOS, doesn't expand `XXX` placeholder when there is a suffix. So `mktemp "$TMPDIR/test-XXXXXX.sh"` creates a literal file `$TMPDIR/test-XXXXXX.sh`, which is fine if trap works and things get removed, but if the process gets killed the file stays and all subsequent runs fail silently


umputun (approved):
confirmed the issue, macOS `mktemp` indeed treats the X's as literal when there's a suffix after them. all sessions end up sharing the same `revdiff-launch-XXXXXX.sh` filename, which is a real problem for concurrent runs.

LGTM, thx

PR #112, opened 2026-04-15
Add Jujutsu (jj) support

nvahalik:
## Summary

- Adds Jujutsu as a third VCS backend alongside git and hg. `DetectVCS` checks `.jj` before `.git` so colocated jj+git repos resolve as jj — reads go through the jj working-copy model instead of bypassing it via git.
- `Jj` renderer translates git-style refs (`HEAD` → `@-`, `HEAD~N` → `@` plus N+1 dashes, `A..B` → `--from/--to`) to jj revsets, runs `jj diff --git --context=1000000`, and feeds the result through the existing `parseUnifiedDiff` pipeline. jj emits raw bytes for binary files instead of the `Binary files … differ` marker, so a pre-parse pass rewrites such hunks to re-use the existing binary placeholder path.
- `--all-files` now supports jj via `NewJjDirectoryReader` backed by `jj file list`. `DirectoryReader` became lister-pluggable (git remains the default; Mercurial is still unsupported for `--all-files`).
- `FileBlame` uses `jj file annotate -T <tab-separated-template>` mirroring the hg blame pattern.
- `--staged` warns and ignores on jj (no staging area), matching hg behavior.

## Test plan

- [x] `go test ./...` passes (race + coverage intact)
- [x] `go vet ./...` clean
- [x] `golangci-lint run` — jj files clean (3 pre-existing `exhaustive` warnings in `app/ui/model.go`/`overlay/annotlist.go` remain, untouched by this PR)
- [x] Unit + e2e jj tests cover `DetectVCS`, ref translation, diff range flag building, rename summary parsing, binary synthesis, blame template parsing, `UntrackedFiles`, `--all-files` via `jj file list`
- [x] Renderer setup tests cover default / `--only` / `--all-files` / `--include` / `--exclude`
- [ ] Manual smoke test against a real colocated jj+git repo

🤖 Generated with [Claude Code](https://claude.com/claude-code)


nvahalik:
First off, I can't thank you enough for this! I happened to stumble across it while looking for something that does this and this seemed like a great starting off point.

I've made these changes and have been playing around with it now locally for doing diff and something-akin-to-interdiff for my own purposes.

I suppose the biggest changes are that JJ's revsets are now supported as an option (i.e. `revdiff <rev>` works) and that JJ's blame support now also works.

umputun (commented):
the Go code looks solid, clean implementation following git/hg patterns. tests are good, binary diff synthesis is a nice catch. couple of things before merge:

1. **rebase needed** - branch is behind master (we merged a structural refactor, dep bumps, and mktemp fix since). the dep downgrades and stale switch statement changes in `model.go`/`annotlist.go` will resolve on rebase

2. **`runJj` duplicates `runVCS`** in `jj.go:274`. the only difference is `extraArgs` prepending. could prepend to args and call `runVCS(j.workDir, "jj", fullArgs...)` instead

3. **skill/plugin not jj-aware** - the Claude Code skill (`detect-ref.sh`) is entirely git-based. in a jj repo every `git` command in that script fails. this doesn't have to be in this PR, but I'd like to track it. could you open an issue for jj support in `detect-ref.sh`?

4. **history module** - `history.Service.gitCommitHash()` returns empty for jj repos since `gitRoot` is empty. not a blocker, just worth a comment somewhere

the `.jj`-before-`.git` detection order makes sense for colocated repos. I checked and jj does need its own backend, `git diff` gives wrong results in jj repos due to the detached HEAD / working-copy-as-commit model

nvahalik:
Re: #4 (history module) — leaving this as-is for now, but here's what a fix would look like:

`gitCommitHash()`, `gitDiff()`, and `filterRepoFiles()` all shell out to git directly. For jj support, the cleanest path would be adding `VCSType` to `history.Params` and dispatching per-method:

| method | git (current) | jj equivalent |
|--------|--------------|---------------|
| `gitCommitHash` | `git rev-parse --short HEAD` | `jj log -r @ --no-graph -T 'commit_id.short(8)'` |
| `gitDiff` | `git diff` / `git diff --cached` | `jj diff` |
| `filterRepoFiles` | `filepath.Rel(gitRoot, ...)` | same logic, just needs a non-empty root |

The `GitRoot` field in `Params` would become `VCSRoot` (or just get populated for jj repos too — the caller in `history_save.go` already has access to the workdir from VCS detection).

umputun (approved):
runJj refactor in 9cd28f8 looks clean, thx. tests pass, lint clean.

LGTM, merging. a few minor doc nits I'll clean up in a follow-up commit:

1. `jjblame.go:61` - "5-way Split" comment is wrong, `SplitN(row, "\t", 4)` produces 4 fields (Copilot A)
2. `jjblame.go:34` - `blameTargetRef` godoc misdescribes the mechanism. code is correct: empty/single-sided ranges return `""` so `FileBlame` omits `-r` and jj defaults to working copy (Copilot C)
3. `directory.go:25` - `NewDirectoryReader` godoc still says "must be inside a git repository", stale now that the lister is pluggable

on Copilot B (binary stdout buffering in `jj.go:147`) - real issue, `runVCS` uses `cmd.Output()` so a multi-MB binary gets buffered just to synthesize a one-line marker. not a blocker, will track as a follow-up issue.

thx for the jj support!

Issue #122, opened 2026-04-18
'i' commit overlay may show different commits than the current diff when HEAD advances during a session


rashpile:
When revdiff is opened against a moving ref (e.g. HEAD~3..HEAD) and new commits are pushed to the branch during the session, the commit info overlay (i) and the file diff can become out of sync:

  1. Open revdiff — file list loaded from the current HEAD
  2. New commit is added externally (HEAD advances)
  3. Press i — commits are fetched lazily now, against the new HEAD
  4. The overlay shows the new commit, but the diff pane still reflects the old HEAD's files

  **Root cause**: the file list is loaded eagerly at startup; commits are loaded lazily on the first i press. If HEAD advances between those two points, they query different VCS states and show inconsistent information.

  **Expected behavior**: commit overlay and file list should reflect the same VCS state. Options: fetch commits eagerly at startup alongside the file list, or show a hint indicating the data may be stale if HEAD has moved since startup.

PR #123, opened 2026-04-18
Add `R` keybinding to reload diff from VCS

rashpile:
## Use case

When using revdiff alongside an iterative workflow — e.g. running ralphex against a base branch while it commits changes — the diff can go stale mid-session. Previously the only option was to quit and restart revdiff with the same parameters. This PR adds `R` to reload the file list and diff in place without leaving the session.

## What's new

- **`R`** — reload the file list and current file diff from VCS. File tree position and view settings (wrap, line numbers, word-diff, etc.) are preserved across the reload.
- **Annotation safety** — if annotations exist, `R` prompts for confirmation first: status bar shows `"Annotations will be dropped — press y to confirm, any other key to cancel"`. Annotations are dropped on confirm since line numbers may shift after a reload.
- **Filter reset** — if the file tree filter (annotated-files-only view) is active when annotations are cleared, the filter is turned off automatically.
- **Commit info sync** — the commit info overlay (`i`) cache is invalidated on reload so the next `i` press re-fetches commits in sync with the new file list.
- **stdin guard** — `R` is a no-op in `--stdin` mode (stream already consumed); a status bar hint explains why.

## Implementation notes

- Reload re-uses the existing two-phase async load pipeline (`loadFiles` → `handleFilesLoaded` → `loadFileDiff`). No new async machinery.
- File position restored via the existing `SelectByPath` call in `handleFilesLoaded`.
- `ReloadApplicable` follows the same composition-root pattern as `CommitsApplicable` — computed in `main.go`, copied into model state, never re-derived from CLI flags inside the model.
- `triggerReload()` is named to avoid shadowing the `reload reloadState` field on `Model`.

## Incidental fix included

- **`handleFilterToggle` stuck-filter fix** — the "filter annotated files only" view (`f` key) previously could not be toggled off once all annotations were deleted: the guard `len(annotated) > 0` skipped the toggle when the store was empty, leaving the user with an empty filtered view. Added `|| m.tree.FilterActive()` to also allow toggling when the filter is already on. Pre-existing bug, surfaced while wiring reload's annotation-drop path. Regression test: `TestModel_HandleFilterToggle_TurnsOffWhenNoAnnotations`.


umputun (commented):
Nice feature, and good that the iterative fix commits already cleaned up the first-round review findings. Tests, race, and linter all clean.

A few things before merge:

1. **docs sync miss** — `plugins/codex/skills/revdiff/references/usage.md` is missing the new `R` row. `.claude-plugin/skills/revdiff/references/usage.md` was updated but the codex copy diverges. CLAUDE.md calls these out as needing to stay byte-identical. `diff` shows exactly one missing line.

2. **stale `file.loadSeq` on reload** — `triggerReload` (loaders.go:109) bumps `filesLoadSeq` but not `file.loadSeq`. Between `R` and the new `filesLoadedMsg` arriving, an in-flight `fileLoadedMsg` from the previous selection still matches `m.file.loadSeq` and gets accepted, briefly overwriting `file.lines` with pre-reload content before `handleFilesLoaded` corrects it (it bumps `file.loadSeq++` at line 163). Self-corrects but creates a flash. One-liner to also bump `m.file.loadSeq` in `triggerReload` closes the window.

3. **duplicated cleanup** — `store.Clear()` + `tree.ToggleFilter(nil)` appears in both `handleReload` (handlers.go:786-805) and `handlePendingReload`. A small private method `applyReloadCleanup()` would dedupe and match the project's method-over-standalone-helper rule.

4. **test gap for commit-cache invalidation** — no regression test locking in the `commits.loaded = false` / `commits.list = nil` reset in `triggerReload` (the #122 fix). A future refactor could silently remove it. One subtest in `TestModel_TriggerReload_...` asserting `commits.loaded` flips to false closes this.

minor: the `handleFilterToggle` `|| m.tree.FilterActive()` change (handlers.go:181) is a legit pre-existing bug fix (stuck filter after deleting last annotation) but isn't called out in the description. Fine to keep, but worth a mention in the PR body or a dedicated test.

rashpile:
fixed address reload PR review findings
- sync R keybinding row into codex skills usage doc
- bump file.loadSeq in triggerReload to drop stale fileLoadedMsg
- extract applyReloadCleanup to dedup handleReload/handlePendingReload
- add regression test for commit cache invalidation on reload
- add regression test for filter-toggle fix
- added PR mentioned about filter fix

umputun (approved):
all 4 findings addressed cleanly, and the filter-toggle fix got its own regression test too. tested the binary locally, `R` works as expected.

tests/lint/race all green. LGTM.

PR #136, opened 2026-04-21
feat: add kitty ssh support

kovstas:
When launching revdiff via kitty in an SSH session, --cwd="$CWD" caused the command to run locally instead of on the remote, producing an empty diff. Replace with --cwd=current (triggers kitty's SSH re-routing) and prepend an explicit cd "$CWD" to the sh -c payload to ensure the correct working directory propagates to the subprocess.


umputun (approved):
lgtm. Copilot's inline comment on `launch-plan-review.sh:93` is a false positive — you added `CWD="$(pwd)"` at line 28 in the same PR, Copilot only looked at the narrow hunk and missed it.

PR #155, opened 2026-04-27
feat: review info popup (i) with --description and aggregate stats

### LINKED ISSUE #130: Proposal: Add a `--description` cli flag so the agent can explain the diff to the user in a dedicated pane.

### LINKED ISSUE #154: feat: review info panel
I often get back to a couple of revdiffs that were opened automatically for me by a workflow. It would be helpful to quickly get additional context on what I am looking at.

A simple info panel like this could be handy in this situation:

Image

melonamin:
## Summary
- Adds a unified review-info popup on `i` showing invocation scope, file/status histogram, aggregate `+/-` line stats, optional markdown description, and the commit log.
- Adds `--description` / `--description-file` CLI flags so an agent can explain the diff inside the popup; file reads are size-capped, regular-file-only, and ANSI-sanitized.
- Pre-rename `commit_info` keybindings keep working via a deprecated alias that rewrites to the canonical `info` action with a one-time WARN.

Closes #154
Closes #130

## Test plan
- [x] `make test` (race + coverage) passes
- [x] `make lint` passes
- [x] Open `i` against a git/hg/jj repo and confirm scope rows, file histogram, line stats, and commit log render correctly
- [x] Pass `--description "**hello**"` and `--description-file path/to/notes.md` and verify markdown highlights inside the popup
- [x] Add `map i commit_info` to `~/.config/revdiff/keybindings` and confirm the deprecation warning + that `i` still opens the popup
- [x] `R` reload during an open popup does not surface stale stats


umputun (changes requested):
tests / lint / race all green, security clean, sanitization layers and the lazy-stats sequencing are well thought out. before merging I'd like a few cleanups, mostly project-convention items:

**test file organization** (per `~/.claude/rules/testing.md`, "one test file per source file"):

1. `app/reviewinfo_test.go` (package main) tests `reviewInfoFromOptions`, `resolveDescription`, `maxDescriptionFileSize` — these live in `app/main.go`. extract them out of `main.go` (which already grew +98) into a new `app/reviewinfo.go` (package main). matches the existing `themes.go`+`themes_test.go`, `stdin.go`+`stdin_test.go`, `config.go`+`config_test.go` shape.
2. `app/ui/reviewinfo.go` (354 lines) has all its UI-side tests piled into `app/ui/loaders_test.go` (~13 ReviewStats tests, plus header/footer/lines text). split them into a new `app/ui/reviewinfo_test.go`.

**methods vs standalone functions** (project rule: functions called only from methods MUST be methods, prefer methods over standalone helpers):

- `app/diff/hg.go:203` `translateRef()` — should be `*Hg` method (`app/diff/jj.go:256` does this correctly with the same shape)
- `app/ui/overlay/overlay.go:400` `injectBorderEdgeText()` — `*Manager` method
- `app/ui/reviewinfo.go:349` `reviewListFlag()` — `Model` method
- `app/ui/overlay/info.go:507/522/534/554` `splitLines` / `padPlain` / `truncateForDisplay` / `sanitizeInfoText` — all called only from `*infoOverlay`, should be methods
- `app/keymap/keymap.go:434` `dumpKeyName()` — `*Keymap` method (already recursive-ready)

**test-only code in production types** (per project preference against test-only fields):

- `ReviewInfoConfig.Enabled` (`app/ui/model.go:348`) — doc-commented as a test-only off-switch but baked into a production type with `m.review.cfg.Enabled` checks scattered across `reviewinfo.go`. clean shape: `ReviewInfo *ReviewInfoConfig` (nil = off), thread nil-checks instead. small surface now, grows otherwise.
- `commitsState.hint` (`app/ui/model.go:331`) — doc admits "no longer set" in production, only tests write to it. either remove the field or remove the test writes.

**parameter-count smells** (4+ → option struct, project rule):

- `app/review/stats.go:53` `ComputeStats(differ, ref, staged, workDir, entries)` — 5 params
- `app/main.go:277` `reviewInfoFromOptions(opts, workDir, vcsType, description)` — 4 params, awkward "options + extras" mix
- `app/review/stats.go:101` `readUntracked(workDir, realWorkDir, relPath, prevPartial)` — 4 params, three same-typed strings (silent-swap risk)
- `app/ui/overlay/overlay.go:381,390` `injectBorderTitle` / `injectBorderFooter` — 5 params each
- `app/ui/overlay/info.go:214` `renderDetailRow(label, value, labelW, innerWidth, muted, reset)` — 6 params with two same-type pairs (`label,value`, `muted,reset`), reorder risk

**unnecessary export**:

- `app/ui/overlay/info.go:514` `NormalizeNewlines` — doc-comment claims external use, but `rg overlay.NormalizeNewlines` returns nothing. unexport.

**deprecated-alias WARN messaging**:

PR body says "one-time WARN" but `app/keymap/keymap.go:546` actually logs once per *occurrence* of `commit_info` in the keybindings file. either track logged-once across program lifetime (small `loggedAliases` map on the parser scope), or relax the wording in the PR description and the alias doc-comment to "logged for each occurrence."

**minor style nits**:

- `app/diff/hg.go:49` `hgStatusToFileStatus` has a redundant `hg` prefix when called as `(h *Hg).hgStatusToFileStatus`. sibling `*Jj` uses `statusToFileStatus`. align.
- `app/review/stats_test.go:80` `resolveWorkDir()` test helper duplicates inline production logic at `stats.go:60-64`. extract a small package-level helper used by both — the doc-comment already acknowledges the duplication.
- `app/ui/reviewinfo.go:162` `highlightDescription()` is a one-line accessor returning a public-ish field that's already read directly elsewhere. inline it.

none of these are blocking individually, but the test-organization items are repeats of the project's stated convention and the methods-vs-standalone batch is a list of seven that have a single clear rule. happy to merge once those are addressed.

umputun (approved):
all 17 items from the prior round addressed, tests / lint / race green.

re Copilot's 4 inline comments: each describes an issue this commit already fixed (Copilot anchored to a stale snapshot). Nothing actionable there.

LGTM.

PR #156, opened 2026-04-27
feat(annotations): --annotations flag preloads annotations from FormatOutput markdown

### LINKED ISSUE #153: Feature request: --annotation flag to preload annotations (LLM review workflow)
## Problem

A common step in my workflow — and in many other vibe-coders' workflows — is asking an LLM to do a code review. The result is a list of comments with references to specific files and lines. To judge whether those comments are actually meaningful, I need to look at the corresponding code in my editor, comment by comment. It would be much nicer to open revdiff with that list of comments already preloaded, where I can edit them, drop the ones that don't make sense, and add my own on top — all in the same view as the diff itself.

revdiff already has the perfect surface for this (annotation gutter, `@` popup, filter-by-annotated tree, hunk navigation); it just has no way to seed annotations from an external source.

## Proposal

Add a repeatable `--annotation <json>` flag that pre-populates the annotation store at startup with comments supplied inline as JSON. Each value is either a single annotation object or a JSON array of them:

```
revdiff --annotation '{"file":"app/main.go","line":42,"type":"+","comment":"check error here"}' \
        --annotation '[{"file":"app/main.go","line":80,"end_line":95,"type":"+","comment":"split this"},
                       {"file":"app/main.go","comment":"file-level note"}]' \
        HEAD~1
```

tushkanin:
Closes #153.

## Summary

Add a `--annotations=PATH` flag that preloads the annotation store from a markdown file in the same shape `Store.FormatOutput` emits. The format is bidirectional: any file written by `-o` can be read back via `--annotations`, so an LLM-generated review (or hand-authored notes, or a previous session) can be opened directly in the TUI for inline curation.

Per the issue discussion, this lands as a single string flag (not repeatable JSON), regex-strict parser over the existing markdown format, and warn-and-drop policy for orphan records.

## What changed

- `app/annotation/parse.go` — new `Parse` for the four header shapes (`(+)`, `(-)`, `( )`, `(file-level)`), including range hunks (`:N-M`), multi-line bodies, and the inverse of `escapeHeaderLines`. Symmetric escape/strip rule so any indent depth round-trips losslessly.
- `app/annotations_load.go` — preload pipeline. Drops orphans against the resolved diff with an stderr warning (file-level needs file-in-diff, line-scoped needs file + matching DiffLine for the record's change type). Untracked files (which surface in the UI via the show-untracked toggle, not `ChangedFiles`) are folded in via `untrackedFn` and validated by reading from disk via `diff.ReadFileAsAdded`, mirroring `Model.resolveEmptyDiff`. Last-write-wins via existing `Store.Add` covers duplicates and preload-then-edit.
- `app/config.go` + `app/main.go` — `--annotations` (no-ini, per-invocation), wired before model construction.
- Docs — README, site/docs.html, CHANGELOG, and the Claude/codex plugin SKILL/usage references get a Preloading Annotations section and a round-trip example.
- `docs/plans/completed/2026-04-27-preload-annotations.md` — implementation plan archived.

## Test plan

- [x] `go test ./...` (full suite passes; new tests in `app/annotation/parse_test.go` and `app/annotations_load_test.go`)
- [x] `golangci-lint run ./app/...` (0 issues)
- [x] Round-trip: `revdiff -o review.md HEAD~1`, edit externally, `revdiff --annotations=review.md HEAD~1` — preloaded state matches saved state
- [x] Orphan drop: annotations referencing files/lines not in the current diff produce stderr warnings and are skipped
- [x] Untracked files: annotations saved against an untracked path round-trip through `--annotations` (regression test `TestPreloadAnnotations_UntrackedFile`)
- [x] Hand-authored bodies with indented `## ` lines round-trip losslessly

### DIFF
--- a/.claude-plugin/skills/revdiff/SKILL.md
+++ b/.claude-plugin/skills/revdiff/SKILL.md
@@ -1,6 +1,6 @@
 ---
 name: revdiff
-description: Review diffs, files, and documents with inline annotations in a TUI overlay, or answer questions about revdiff usage, configuration, themes, and keybindings. Opens revdiff in tmux/zellij/kitty/wezterm/cmux/ghostty/iterm2/emacs-vterm, captures annotations, and addresses them. Works in git, hg, and jj repos (auto-detected). Activates on "revdiff", "review diff", "review changes", "annotate diff", "git review with revdiff", "hg review with revdiff", "review jj change", "interactive diff review", "revdiff all files", "review all files", "browse all files", "revdiff <file>", "revdiff README.md", "revdiff /tmp/notes.txt", "review this file", "annotate this file", "review file with revdiff", "revdiff config", "revdiff themes", "revdiff keybindings", "how to configure revdiff", "what themes does revdiff have".
+description: Review diffs, files, and documents with inline annotations in a TUI overlay, or answer questions about revdiff usage, configuration, themes, and keybindings. Opens revdiff in tmux/zellij/kitty/wezterm/cmux/ghostty/iterm2/emacs-vterm, captures annotations, and addresses them. Works in git, hg, and jj repos (auto-detected). Activates on "revdiff", "review diff", "review changes", "annotate diff", "git review with revdiff", "hg review with revdiff", "review jj change", "interactive diff review", "revdiff all files", "review all files", "browse all files", "revdiff <file>", "revdiff README.md", "revdiff /tmp/notes.txt", "review this file", "annotate this file", "review file with revdiff", "open this review in revdiff", "show review in revdiff", "review in revdiff", "revdiff config", "revdiff themes", "revdiff keybindings", "how to configure revdiff", "what themes does revdiff have".
 argument-hint: 'optional: ref(s), "all files", or file path'
 allowed-tools: [Bash, Read, Edit, Write, Grep, Glob]
 ---
@@ -18,6 +18,7 @@ Review diffs with inline annotations using revdiff TUI in a terminal overlay. Wo
 - "revdiff all files exclude vendor"
 - "revdiff README.md", "revdiff docs/plan.md", "revdiff /tmp/notes.txt" — single-file review (`--only` mode)
 - "review this file", "annotate this file", "review file with revdiff"
+- "open this review in revdiff", "show review in revdiff", "review in revdiff" — open an in-session review (preload mode)
 
 ## Answering Questions
 
@@ -36,6 +37,10 @@ ${CLAUDE_SKILL_DIR}/scripts/read-latest-history.sh
 ```
 
 The script resolves the history dir from `$REVDIFF_HISTORY_DIR` (default `~/.config/revdiff/history`), finds the repo subdir via VCS root basename (jj/git/hg), and prints the newest `.md` file found. Each history file contains a header (path, refs, and — when available — a git commit hash), the annotations in `## file:line (type)` format, and the raw git diff for annotated files. The `commit:` line and diff block are captured from git only; in hg/jj repos the diff block will be empty and no commit hash is recorded. See `references/usage.md` "Review History" section for directory layout, stdin/only handling, and override options.
+
+## Opening an In-Session Review
+
+When the user asks to open an in-session review in revdiff (the conversation already contains review comments produced earlier in the session), write those comments to a temp file (e.g. `/tmp/revdiff-review-XXXXXX.md`) using the format documented in `references/usage.md` ("Output Format" section), then run the normal launcher flow (Step 1 ref detection, Step 2 invocation) with `--annotations=<temp-path>` appended. Step 3 onward handles the curated annotations as usual.
 
 ## How It Works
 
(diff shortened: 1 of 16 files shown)


umputun (changes requested):
tests / lint / race green. parser tests are well shaped: table-driven, round-trip property test, path-with-colon edge case all covered. before merge a few things worth addressing:

**block:**

1. **sanitize comment text on parse.** `Annotation.Comment` flows verbatim from file into `Store.Add`, then through `Renderer.AnnotationInline` which wraps raw bytes in ANSI italic. The PR's headline use case is LLM-generated `review.md`, and the new skill flow in `SKILL.md` ("Opening an In-Session Review") makes Claude itself the agent producing the file from arbitrary conversation context. That context can plausibly contain raw `\x1b[...]` sequences, CR-overwrite tricks, or stray escapes (terminal output pasted into chat, error messages with color codes, code samples that discuss ANSI rendering). Without sanitization those bytes reach the TUI render unchanged. The codebase already has `diff.sanitizeCommitText` for exactly this threat model (commit author/subject/body); route every parsed `Comment` through it before `store.Add`. About 5 lines. If #155 lands first, just import the now-exported `diff.SanitizeCommitText`. If this lands first, export the symbol here or duplicate the strip table.

2. **harden file read to match #155.** `app/annotations_load.go:26` is naked `os.Open(path)` with `nolint:gosec`. Mirror `app/main.go:resolveDescription` from #155: `os.Stat` first, reject non-regular files via `IsRegular()` (otherwise opening a FIFO blocks the process before bubbletea initializes), then `os.Open`, then `io.LimitReader(f, max+1)` for an upper bound. 1 MiB cap is reasonable here since annotation files aggregate many records. Less critical for the skill-driven temp-file path since the file is skill-controlled, but matters for curl'd review files and edited history files.

**should fix:**

3. **extract a `preloader` struct, fix params and standalones in one move.** the wide-signature problem and the standalone-helpers problem are the same problem.
   - `preloadAnnotations` (8 params), `lookupLineSet` (8 params), `resolveKnownFiles` (5 params), `buildLineSet`. Project rule is 4+ → option struct.
   - all four are called only from each other (single caller for each helper). Project rule is "functions called only from methods MUST be methods, prefer methods over standalone helpers."
   - one refactor closes both: `type preloader struct { renderer, ref, staged, untrackedFn, workDir, warnOut, lineCache }`, then `(p *preloader) load`, `(p *preloader) resolveKnownFiles`, `(p *preloader) lookupLineSet`, `(p *preloader) buildLineSet`. tests collapse from `"", false, nil, "", &bytes.Buffer{}` repetition to a single `preloader{...}` literal.

   `parseHeader` in `app/annotation/parse.go:99` is the same shape (standalone, single caller `Parse`). If `Parse` becomes `Store.Load` per #4, `parseHeader` naturally becomes a method on whatever parser type that introduces, or stays as an unexported helper local to `Load`.

**discuss:**

4. **`Parse` → `Store.Load` / `Store.Merge`.** Right now `Parse` returns `[]Annotation` and `preloadAnnotations` re-feeds each through `store.Add`. A method on `Store` would close the round-trip loop with `Store.FormatOutput` symmetrically on the API surface, not just on the format level, and remove the for-loop in the loader.

5. **skill instruction underdocs the body-escape rule.** `SKILL.md` tells Claude to write the temp file "using the format documented in references/usage.md (Output Format section)". That section has the `## ` → ` ## ` body-escape rule (line 245), but it's tucked into a paragraph about output. When Claude generates fresh content for `--annotations`, it can easily emit a body line starting with `## ` without escaping, which the strict parser rejects (or worse, silently splits the record). Either add an explicit "escape `## ` in body lines" callout next to the skill instruction, or relax the parser to tolerate unescaped `## ` in non-header positions. A regression test covering "body line starts with unescaped `## `" would also help.

6. **renamed-file silent drop.** If a file was renamed since `--annotations` was generated, records key on the old path and get orphan-dropped. Known limit of the format keyed on path. A one-line note in `resolveKnownFiles` doc-comment saves the next reader from chasing it as a bug.

7. **`--stdin --annotations` is undefined.** Preload runs after stdin payload setup so file-level records keyed on the synthetic stdin name probably work; line-scoped records orphan-drop. Either add to the stdin mutual-exclusion list in `parseArgs` or document as supported.

**nits:**

- `app/annotations_load.go:48,57,85,99,142`, five occurrences of `_, _ = fmt.Fprintf(warnOut, ...)`. Once the option-struct refactor lands, a `(p *preloader) warnf(...)` helper collapses those.
- `app/annotations_load.go:81-93`, doc-comment says "matches `ui.loadFiles`" but `loadFiles` only folds untracked when `showUntracked` is on; preload always folds. Clarify the divergence.

umputun (approved):
all addressed cleanly: sanitization wired through `diff.SanitizeCommitText`, file-read hardened with Stat→IsRegular→LimitReader and 1 MiB cap, preloader struct holds the state, `Store.Load` closes the round-trip, parser tolerates unescaped `## ` after the first header (with regression test), `--stdin --annotations` rejected, doc-comments updated.

verified locally with the built binary against rejection paths (FIFO, dir, oversize, malformed, --stdin mutex), orphan-drop warning, and the unescaped `## ` tolerance. tests / lint / race green.

LGTM.

PR #157, opened 2026-04-29
Add colorblind-friendly gallery themes

krajcik:
## Problem

revdiff has several bundled and gallery themes, but no theme pair aimed at users who have difficulty relying on red/green distinctions in diffs.

## Solution

Add two gallery-only community themes:

- colorblind-dark
- colorblind-light

Both use high-contrast add/remove colors and existing Chroma styles. They do not change defaults and are not marked as bundled.

## Testing

- GOTOOLCHAIN=auto make validate-themes
- Manually previewed colorblind-dark and colorblind-light in WezTerm using both light and dark terminal color schemes
Снимок экрана 2026-04-29 в 13 57 18
Снимок экрана 2026-04-29 в 14 00 11

### DIFF
--- a/themes/gallery/colorblind-dark
+++ b/themes/gallery/colorblind-dark
@@ -1,0 +1,26 @@
+# name: colorblind-dark
+# description: high-contrast red-green colorblind-friendly dark palette adapted from GNU Modus
+# author: Nikita Krajcik
+
+chroma-style = modus-vivendi
+color-accent = #2fafff
+color-border = #646464
+color-normal = #ffffff
+color-muted = #989898
+color-selected-fg = #f0f0f0
+color-selected-bg = #2a2a6a
+color-annotation = #ffa00f
+color-cursor-fg = #efef00
+color-cursor-bg = #1a1a1a
+color-add-fg = #c4d5ff
+color-add-bg = #003066
+color-remove-fg = #d4d48f
+color-remove-bg = #3d3d00
+color-modify-fg = #e3cfff
+color-modify-bg = #2f123f
+color-tree-bg = #000000
+color-diff-bg = #1a1a1a
+color-status-fg = #f0f0f0
+color-status-bg = #2a2a6a
+color-search-fg = #000000
+color-search-bg = #efef00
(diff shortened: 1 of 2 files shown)


umputun (changes requested):
I tested both themes locally in kitty. Dark looks fine.

Light has a real issue: `chroma-style = xcode` uses `LiteralString = #c41a16` (red) and `Comment = #177500` (green), so syntax highlighting in viewed files defeats the whole point of the theme. Very visible on markdown content.

modus-vivendi (which dark already uses) has a matching light counterpart `modus-operandi` in chroma. Strings blue (`#2544bb`), comments grey (`#505050`), no red/green pair. That's the right pick for the light variant since the description already cites GNU Modus.

other than that lgtm, thx for the contribution.

upd: this is what light looks like in my kitty:

image

umputun:
to expand on the chroma-style point: it's worse than just red/green tokens. Xcode's `Background` entry sets only a bg (`bg:#ffffff`) and no fg, while modus-vivendi sets both (`#ffffff bg:#000000`). Most markdown tokens (plain text, list items, headers via `GenericHeading`) have no explicit entry in xcode, so chroma's inheritance chain ends at `Background` with no fg color.

revdiff's `writeTokenANSI` (`app/highlight/highlight.go:147`) only writes an ANSI fg when chroma sets one, and `contextHighlightStyle` (`app/ui/style/resolver.go:406`) deliberately omits fg expecting chroma to provide it. Net effect: most CHANGELOG/markdown content renders with no fg at all and falls through to the terminal's default fg, which on a light bg is invisible.

modus-operandi is symmetric to modus-vivendi (`#000000 bg:#ffffff`), so unmatched tokens get explicit black fg through the same fallback chain. That's why dark works and light doesn't.

krajcik:
Done, thanks. Now looks good. Fun fact: I'm colorblind myself and didn't realize the text was red.

umputun (approved):
lgtm, thx. The faint `+`/`-` prefix on highlighted lines is a revdiff rendering bug, not on your theme. The highlighted line style at `app/ui/style/resolver.go:193` sets bg only and expects chroma to set fg per token, but the prefix is prepended outside the chroma path so it falls through to the terminal default fg. I'll patch that separately.

PR #163, opened 2026-05-01
feat: `--compare=old:new` — two-file diff for rolling agent reviews

rashpile:
## Motivation

revdiff is built for **rolling review loops with a Claude agent**: the agent produces content, the user annotates lines, the agent revises, the loop repeats until the user quits without annotations. This works well today for code diffs, where each iteration is naturally framed as a `+`/`-` view of what the agent just changed.

The same loop falls apart for **agent-rewritten documents** — plans, design docs, specs, generated reports. Today the agent can only re-open the rewritten file as context-only (`--only`), so the user has to re-read the whole document to find what changed in response to their last round of annotations. Long documents become exhausting to review iteratively, and small revisions are easy to miss.

`--compare=old:new` fixes this. The agent snapshots the document before rewriting it, then re-opens revdiff against the snapshot:

```
agent shows v1     → user annotates "tighten section 3, drop section 5"
agent rewrites     → snapshots v1, opens revdiff --compare=v1:v2
user sees only what changed → annotates again, or quits
```

Each iteration becomes a focused diff of *just this round's edits*, with full annotation support — the same loop revdiff already supports for code, now extended to any document the agent revises. The missing primitive for iterative agent-driven document review.

## Usage

```bash
revdiff --compare=/tmp/plan-v1.md:docs/plans/plan.md
```

`--compare` is mutually exclusive with: refs, `--staged`, `--only`, `--all-files`, `--stdin`, `--include`, `--exclude`, `--annotations`.

## How it works

`CompareReader` runs `git diff --no-index` and feeds the output through the same `parseUnifiedDiff` parser used by every other diff source. Because the display layer is unchanged, every existing feature works automatically: word-diff, compact mode with correct line-count dividers, syntax highlighting, scrollbar, and inline annotations.

A few non-obvious details:

- Exit code 1 from git (files differ — the normal case) is treated as success. Exit 1 with empty stdout is a real failure (e.g. missing file).
- In compact mode, the old file's line count is needed for trailing dividers. It's computed via streaming reads, not `os.ReadFile`, so large files don't get pulled into memory.
- No git repository is required — `git diff --no-index` works anywhere `git` is on `$PATH`.

## Changes

- **`app/diff/compare.go`** — new `CompareReader` type implementing `Renderer`
- **`app/config.go`** — `--compare` flag + `validateCompareFlag` (parses `old:new`, validates both paths are regular files, enforces mutual exclusions)
- **`app/main.go`** — wires compare mode before stdin/VCS dispatch; `resolveComparePaths` resolves both paths to absolute; `reloadApplicable` returns true for compare mode
- **`app/reviewinfo.go`** — info overlay shows "two-file diff" header for compare mode
- **README, site/docs.html, plugin usage reference** — documented with examples

## Verification

```bash
make test                                         # unit tests pass
echo "a" > /tmp/a.txt && echo "b" > /tmp/b.txt
revdiff --compare=/tmp/a.txt:/tmp/b.txt           # two-line diff opens
revdiff --compare=/tmp/a.txt:/tmp/a.txt           # identical files: empty diff, no crash
revdiff --compare=/tmp/a.txt:/tmp/b.txt --stdin   # error: mutual exclusion
```

## Out of scope

- A planning skill / hook that orchestrates this loop automatically (snapshot before rewrite, invoke `--compare` after) is **not** part of this PR. That's a separate design conversation; this PR ships the primitive.


umputun (changes requested):
recommendation: scrap `--compare` and instead extend `--stdin` to parse unified-diff input (`--stdin-format=patch` or auto-detect `diff --git` headers). That covers `git diff --no-index a b | revdiff --stdin --stdin-format=patch` in maybe 30 lines vs ~290 here. A calling skill wraps either invocation just as cleanly so launch friction doesn't matter, and the remaining differences (info popup label, both-paths display) are cosmetic and solvable on the stdin side with small extensions. Same workflow, much less surface area, no new top-level mode.

separately: the framing "rolling agent review" oversells what this is. The actual gap is mechanical: revdiff has no clean way to diff two on-disk files when neither side is anchored in a git ref. That's a generic primitive (config backups, upstream-vs-fork comparisons, etc), not specifically agent-shaped.

independent of which approach: plugin docs not synced. `.claude-plugin/skills/revdiff/references/usage.md` updated but `plugins/codex/skills/revdiff/references/usage.md` was not. Per CLAUDE.md they must stay byte-identical.

issues below apply only if you keep `--compare` as proposed; switching to the stdin path would moot most:

1. **colon path parsing is fragile**. `strings.Cut(opts.Compare, ":")` at `app/compare.go:65` splits on first colon. Breaks legal POSIX filenames containing `:` (e.g. timestamped snapshots like `2026-05-01T12:30:00.md`) and forecloses any future Windows port. Zero tests for colon-in-path. Options: keep syntax and document the limit, switch to two repeated flags (`--compare-old=`/`--compare-new=`, matches `--include`/`--exclude`), or two positional args (`revdiff --compare /old /new`, matches `revdiff master feature`). Either alternative kills the bug.

2. **DRY hazard**. `validateCompareFlag` (`app/compare.go:13-58`) and `resolveComparePaths` (`app/compare.go:64-75`) both parse `opts.Compare` with `strings.Cut(":", ...)` separately. Either have validate return `(absOld, absNew, error)` or extract a private `parseCompareFlag` shared by both, otherwise the parser semantics drift.

3. **fragile test**. `TestParseArgs_CompareConflicts` uses non-existent `/tmp/a:/tmp/b` paths and only passes because conflict checks return before `os.Stat`. If check order ever flips to stat-first, every conflict subtest silently starts passing for the wrong reason. Use real tempdir paths or assert both error type and message.

4. **`countFileLines` should be a method on `CompareReader`**. `app/diff/compare.go:79`, called only from `(r *CompareReader).FileDiff`. Project rule: functions called only from struct methods must be methods.

5. **test coverage gaps**. Nothing covers: paths-with-colons (relates to #1), binary files in compare mode, large files (the streaming `countFileLines` exists for this reason and isn't exercised), no-trailing-newline through the full compare flow, exit-code-2 from git, stderr-only success path, symlinks, or `resolveComparePaths` directly.

6. **mutual-exclusion check**. `--compare` is not excluded from `--description`/`--description-file`. Intentional (orthogonal context) or oversight?

minor: insert `"--"` before path args in the argv at `app/diff/compare.go:34`. `filepath.Abs` already makes paths absolute so `-flag-shaped` paths can't reach git in practice, but the `--` hardens that guarantee against future refactors.

tests/lint/race green, argv-style exec safe, no scope creep, no vendor changes.

rashpile:
@umputun My use case for this is a planning hook in the agent loop. I've already validated the approach locally with --compare, but I can adapt to a stdin-based implementation if you'd prefer.

On --compare vs. extending stdin: I'd push back gently on the redesign. revdiff is a diff review tool - "diff two files" feels like a primary operation, not a corner case to compose. The stdin route would also lose compact-mode toggle and reload-on-demand (the bytes are consumed; revdiff doesn't own the source). The 290-line cost is real, but I think the symmetry with existing modes (--staged, --all-files, --only) and the cleaner UX justify a first-class flag rather than a pipe-composed alternative.

Here's the design for plan-review-hook.py:

## Design: plan-content-as-state

The hook is stateless on disk between invocations — Claude Code doesn't persist hook state, and inventing a sidecar registry would be brittle across parallel sessions. Instead, **the plan itself carries its diff history** via an html-comment marker on the first line:

```html
<!-- previous revision: /tmp/plan-rev-AAA.md -->
```

The marker is invisible in rendered markdown, trivially greppable, and self-cleaning (each plan chain references its own snapshot, so parallel sessions can't collide). On every `ExitPlanMode` the hook:

1. Reads the marker from the new plan (if present)
2. Strips it from the content saved to disk — keeps diffs clean
3. Writes the stripped content to a fresh `/tmp/plan-rev-*.md` snapshot
4. Opens revdiff in `--compare` against the marker target if it exists, else falls back to `--only`
5. On annotations → **deny** + tells the agent to put `<!-- previous revision: <new-snapshot> -->` as the first line of the next plan
6. On no annotations → **ask** + deletes the new snapshot (loop is over)
7. Deletes the old snapshot after the compare succeeds


```
agent calls ExitPlanMode (v1)   → hook saves /tmp/plan-rev-AAA.md, opens --only
user annotates "tighten step 3" → hook denies, tells agent to start v2 with marker
agent calls ExitPlanMode (v2)   → hook reads marker, saves /tmp/plan-rev-BBB.md,
                                   opens --compare=AAA:BBB (focused diff of round 2)
user annotates / quits clean    → loop continues or terminates with cleanup
```

Does this plan look OK to you?

umputun:
fair points on compact-mode and orchestrator. Compact toggle re-runs the diff with smaller context, which `--compare` does cleanly via the same constructor; `--stdin` would have to fake it by post-trimming context locally. Real functional differentiator, I had this wrong. Combined with the planning hook actually being in flight (not speculative), `--compare` as a first-class mode is fine. Withdrawing the scrap-and-replace recommendation.

line-level issues from the original review still stand: colon parsing fragility (#1), plugin docs sync miss, DRY hazard, fragile test, `countFileLines` shape, test gaps, mutual-exclusion gap, the `--` argv nit. None of those depend on which mode shape you pick.

on the hook design:

the marker-in-plan approach is clever. It keeps state in band so parallel sessions don't collide and hooks-are-stateless isn't a problem. Two concerns:

1. agent-discipline risk. Chain depends on the agent prepending `<!-- previous revision: ... -->` to v2+. If it forgets (tool drift, context truncation, model swap), the hook treats v2 as v1 and opens `--only` instead of `--compare`. User loses the focused diff for that round. Recoverable since fallback is `--only` not a crash, but the deny-message instruction has to survive across multiple turns. Probably fine in practice, worth acknowledging the failure mode.

2. snapshot cleanup timing. "deletes old snapshot after compare succeeds": which exit counts as success? Revdiff exits non-zero on annotated quit, zero on clean quit. If any exit deletes AAA, a user who closes mid-thought breaks the next iteration. Pin down the trigger explicitly.

design looks good otherwise. Ship it once the line-level items are addressed.

rashpile:
Pushed fixes addressing the review items:

- **Flag shape**: replaced `--compare=old:new` with `--compare-old=<path> --compare-new=<path>`; killed the colon-split parser and `resolveComparePaths` — single parse+validate path, conflict checks before stat checks.
- **Argv hardening**: inserted `--` separator before paths in the `git diff --no-index` invocation.
- **`countFileLines` method-ized**: now `(r *CompareReader) countFileLines()`, drops the redundant path param.
- **Tests**: `TestParseArgs_CompareConflicts` now uses real `t.TempDir()` files; added subcases for paths-with-colons, binary files, large-file streaming (heap smoke-test), no-trailing-newline matrix, git stderr surfacing, symlinks (both + mixed), same-file.
- **`--description` kept allowed**: doesn't conflict; added explicit "see comment" rationale in validator.
- **Docs**: README, site/docs.html, CLAUDE.md, skill usage refs all updated to the two-flag form; added `--annotations` to the documented mutex list (validator already rejected it).

### Added hook implementation — PR description is updated

rashpile:
addressed new findings 

- swap compare-mode arg order from (old, new) to (new, old) in launch-plan-review.sh it simple resolve for custom 1-arg launcher made before compare mode shipped - silently degrade to --only review of the NEW revision instead of opening the OLD one the user already reviewed.

- trusted_snapshot: tighten the deny-reason text so the agent is explicitly told to ignore any other plan-rev-*.md marker it may have seen earlier in the conversation, addressing intra-session chain-mixup at the prompt layer.

PR #174, opened 2026-05-08
feat(markdown): render annotation prose as glow-style markdown

rlch:
Annotation bodies now render as styled markdown — bold, italic, inline code, fenced code blocks (chroma), strikethrough, lists, links, blockquotes. Opt out via `--plain-annotations` / `REVDIFF_PLAIN_ANNOTATIONS`; auto-disabled under `--no-colors`.

Annotation input is multi-line (textarea). Enter saves, Ctrl+J / Alt+Enter insert newline (gum-write convention). Shift+Enter is terminal-dependent — bubbletea v1 can't distinguish it from Enter.

## How it's built

- `app/markdown/` wraps `charmbracelet/glamour` v1.0.0. Foreground-only style (composes with revdiff's pane backgrounds, same rule `app/highlight` follows). `\x1b[0m` post-processed to a fg+attrs reset so outer pane bg survives.
- `annotationVisualRows(prefix, body)` is the single source of truth — both `wrappedAnnotationLineCount` (height) and `renderWrappedAnnotation` (paint) read one cached `[]string`. Cache cleared on file load + theme apply. Markdown renderer rebuilt on theme apply so `AnnotationFg` tracks live.
- Theme integration uses existing color slots: body uses `Annotation`, inline code + links use `Accent` (SearchFg equals DiffBg in every dark theme — would render invisible). No new theme keys.

## Vendor

glamour v1.0.0 + transitive (goldmark, bluemonday, muesli/reflow). ~1.2 MB. lipgloss bumps to v1.1.1-pre (same minor, API-stable).

## Known limitations

- glamour's muesli/reflow occasionally orphans a single short word onto its own line at certain widths. Reproduces in glow itself; shipped as-is.
- Long fenced code blocks in annotations expand to N rows

## Test plan

- [x] `make test` — green
- [x] `go vet ./...` — clean
- [x] Manual fixture at `examples/markdown-annotations.md` (use `revdiff --annotations` against working tree)
- [x] `--plain-annotations` falls back; theme switch updates colors immediately; Ctrl+J grows textarea; Enter trims trailing whitespace before save

image

image


umputun (commented):
three procedural notes first.

CONTRIBUTING.md "first-time contributors" (line 7-9) asks first PRs to start with an issue describing what to build and why, then wait for approval before writing code. That step exists to prevent the situation we have here: 935 LoC production + 820 LoC tests + ~1.2 MB vendor, none of it valuable if the answer is "no". Pls file an issue or discussion for the direction first next time.

commit shape is unreviewable. The bulk of the PR (commit `0a17e86`) is a single commit of 282 files / ~80k insertions mixing vendor with production code, tests, and docs. Copilot refused to review for that exact reason ("exceeds the maximum number of lines (20,000)"), and humans aren't going to do better. Split the next one: separate `go mod vendor` from the actual code change, and split the code change itself if it's large.

discussion #164 ("Feat: markdown preview", 2026-05-01) is directly on point and not referenced in this PR. There I declined glamour-based rendering for `.md` plan-review prose with detailed reasoning about complexity, theme integration, code-fence collision with chroma, reflow vs line-anchoring, and value/cost. I also suggested a narrower "hide markup" alternative.

most of those objections transfer to annotations:

- in my terminal `**bold**` doesn't render as actual bold weight (most terminal fonts render `\x1b[1m` as bright color, not bold). The visible delta is asterisks/backticks hidden, `-` bullets become `•`, fenced blocks framed. That's the feature.
- `lipgloss v1.1.1-0.20250404203927-76690c660834` is a pseudo-version pinned to Charm's master snapshot from April. Maintenance tax until they tag v1.1.1 stable.
- bluemonday is an HTML sanitizer pulled to render content with no HTML.
- `Ctrl+E` external editor already covers multi-line composition. Textarea swap is in-place entry, not new capability.
- chroma already vendors a markdown lexer (`vendor/github.com/alecthomas/chroma/v2/lexers/markdown.go`). Pointing chroma at annotation bodies with the markdown lexer colors markup without hiding it, at zero new vendor and ~50 LoC. That's the "hide markup but keep lines 1:1" idea from #164 applied to annotations.

asymmetry with diff rendering: revdiff intentionally doesn't render markdown for `.md` diff content (also from #164). Drawing the line at "annotations are reviewer prose" is defensible, but `.md` plan reviews are the same prose category and I declined them in #164.

what's worth keeping regardless of the direction:

- `annotationVisualRows` chokepoint + `rowCache` is independent of glamour and fixes a real cursor-math/painter desync class.
- `existingMultiline` workaround removal is real cleanup.

if you want to take another pass:

1. file a discussion describing the use case. Likely it's the agent-injected `--annotations` flow (#153) where bodies really are markdown, strongest case for any prose styling.
2. chroma + markdown lexer as the lighter path. Zero new vendor, no pre-release, no asymmetry.
3. if in-place multi-line entry is the ask, that's a separate smaller PR. Textarea + keymap, no markdown.
4. cherry-pick the `annotationVisualRows` work as its own PR.

declining the markdown direction here. Implementation discipline is good (consumer-side interface, moq mocks, doc sync across all reference copies, rare for first contact), but the strategic call is the same as #164.

Issue #167, opened 2026-05-04
Bundled `pi` skill doesn't include any of the launching scripts


dljsjr:
The Pi extension's bundled skills only describe the slash commands; slash commands are only user-invokable, the agent can't invoke slash commands on behalf of the user, so it blocks interactive Pi use via prompting.

I didn't do a thorough test, but I did do a spot check of the `launch-revdiff.sh` script included with the Claude plugin, and it worked without issue. So I'm thinking that the Pi skill can likely use the existing scripts as-is.

umputun:
analysis is correct. The pi `SKILL.md` only documents the slash commands, and pi slash commands are user-invokable only - so when the agent is prompted with "review my diff", it has no way to actually launch revdiff and falls back to telling the user to run `/revdiff` themselves.

the Claude (`.claude-plugin/skills/revdiff/`) and codex (`plugins/codex/skills/revdiff/`) skills both have the agent-driven shape - SKILL.md instructs the agent to call `launch-revdiff.sh` via Bash. Pi is the odd one out, and since `pi install <repo>` clones the whole repo, the scripts under `.claude-plugin/skills/revdiff/scripts/` are already on disk after install. The gap is just in the SKILL.md content.

worth noting the pi extension at `plugins/pi/extensions/revdiff.ts:24-32` already resolves `.claude-plugin/skills/revdiff/scripts/launch-revdiff.sh` for the slash-command path, so the two paths can coexist (user-side via `/revdiff`, agent-side via Bash) without conflict.

cc @melonamin - you authored the pi integration in #52 and would have the best read on whether the agent-side path should reuse `.claude-plugin/skills/revdiff/scripts/launch-revdiff.sh` directly, or get its own copy under `plugins/pi/skills/revdiff/scripts/`. There's also the override-chain note in CLAUDE.md (pi has no `CLAUDE_PLUGIN_DATA`, so the resolver fallback differs from Claude). @dljsjr is offering to take a swing at it and has already validated the launcher works, would be good to align on the shape before they open a PR.

melonamin:
Here’s my read: the scripts aren’t actually missing from a normal pi install - the full repo is cloned, including .claude-plugin/skills/revdiff/scripts/.

The mismatch is that Pi, unlike codex and claude, has a built-in ability to customize it's interface. Pi's native integration path: `/revdiff` uses `ctx.ui.custom()` to suspend Pi, run revdiff full-screen, then restore Pi and show results in a Pi overlay/widget. 

[...]

PR #182, opened 2026-05-11
Add pi revdiff review agent tool

melonamin:
## Summary
- add a `revdiff_review` pi agent tool for interactive reviews
- update pi extension imports/dependencies and package included files
- document agent-tool usage in the pi skill

## Notes
- `@mariozechner/pi-*` is deprecated upstream and renamed to `@earendil-works/pi-*`; v0.74 is the first release with the agent-tool API.
- The new pi API drops the `success` notify level, so the two completion notifications now use `info`.

Refs #167

### DIFF
--- a/package.json
+++ b/package.json
@@ -14,9 +14,14 @@
     "annotations",
     "tui"
   ],
+  "files": [
+    "plugins/pi/",
+    ".claude-plugin/skills/revdiff/scripts/"
+  ],
   "peerDependencies": {
-    "@mariozechner/pi-coding-agent": "^0.60",
-    "@mariozechner/pi-tui": "^0.60"
+    "@earendil-works/pi-coding-agent": "^0.74.0",
+    "@earendil-works/pi-tui": "^0.74.0",
+    "typebox": "^1.1.24"
   },
   "pi": {
     "extensions": [
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "revdiff-pi",
-  "version": "0.1.2",
+  "version": "0.2.0",
   "description": "Pi package for revdiff interactive diff review",
   "license": "MIT",
   "repository": {
--- a/plugins/pi/README.md
+++ b/plugins/pi/README.md
@@ -4,7 +4,7 @@ This directory contains the **pi-specific** integration for revdiff.
 
 ## Contents
 
-- `extensions/revdiff.ts` — pi extension that launches revdiff, captures annotations, and shows them in pi
+- `extensions/revdiff.ts` — pi extension that launches revdiff, captures annotations, exposes the `revdiff_review` agent tool, and shows results in pi
 - `extensions/revdiff-post-edit.ts` — optional post-edit reminder extension (disabled by default)
 - `skills/revdiff/SKILL.md` — pi skill describing the review workflow and commands
 
--- a/plugins/pi/extensions/revdiff-post-edit.ts
+++ b/plugins/pi/extensions/revdiff-post-edit.ts
@@ -1,4 +1,4 @@
-import type { ExtensionAPI, ExtensionContext } from "@mariozechner/pi-coding-agent";
+import type { ExtensionAPI, ExtensionContext } from "@earendil-works/pi-coding-agent";
 
 const CONFIG_TYPE = "revdiff-reminder-config";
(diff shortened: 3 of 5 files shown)


umputun (commented):
**issues**

1. **false-clean review path** at `plugins/pi/extensions/revdiff.ts:548`. `(exitCode ?? 0) !== 0` treats `undefined` as success, so if `ctx.ui.custom` ever resolves to undefined (pi's degraded RPC mode), this reads no output file and returns an empty `ReviewState`, which the tool then reports as "Review complete - no annotations". The slash command hits the same flaw but the user sees the missing terminal takeover; the agent has no visual feedback and tells the user the diff is clean when revdiff never ran. Suggest gating on `typeof exitCode === "number" && exitCode === 0` AND `existsSync(outputFile)` before treating as success.

2. **version bump** - `package.json` stays at `0.1.2`. CLAUDE.md asks for a bump on any pi plugin change, and the scope here (new public agent-tool API + peer-dep scope migration + new typebox peer-dep) warrants `0.2.0`.

3. **tool returns rawOutput + details** at `revdiff.ts:366`. The annotation content is delivered twice, once in `content` as raw markdown, once in `details` as structured `ReviewState`. Bloats the agent's context on large reviews. Suggest a short headline in `content` (`captured N annotations for X`) with full data only in `details`.

4. **missing `ctx.isIdle()` guard on tool path**. `startReview` guards at line 163; tool's `execute` does not. If the new SDK guarantees `execute` only fires when idle, a one-line comment helps the next reader. Otherwise add the same guard for symmetry.

5. **PR body** - one sentence each on the two bundled changes saves every future reviewer the same investigation:
   - `@mariozechner/pi-*` is deprecated upstream and renamed to `@earendil-works/pi-*`; v0.74 is the first release with the agent-tool API
   - the new pi API drops the `success` notify level, so the two notifications fall back to `info`

**nits**

- `mode` schema at `revdiff.ts:319` - `Type.String(...)` accepts any string and rejects unknown values at runtime via `parseToolMode`. Switching to `Type.Union([Type.Literal("direct"), Type.Literal("overlay")])` surfaces valid values in the tool's schema for the LLM.
- `runAgentReview` at `revdiff.ts:381-392` is a near-trivial wrapper around `runReview` (one progress emit plus a pointless `if (!result) return undefined; return result;`). Could inline at the `execute` callsite.
- `parseToolMode` + caller at `revdiff.ts:317-321` returns `undefined` for both "absent" and "invalid", caller re-checks `params.mode && !mode`. Three outcomes from a two-state return. Inline the validation or return a discriminated result.
- `revdiff:launch` event emitted from `execute` for the tool path but from `startReview` for the slash-command path. No listener distinguishes them today but a future one will. Move into `runReview` once.

melonamin:
sorry for the noise, this one was supposed to be a draft...

melonamin:
Addressed review feedback in 96446ff:

- tightened direct-launch completion handling so missing/undefined exit status or missing output file no longer reports a clean review
- bumped pi package version to 0.2.0
- shortened tool text output and kept full annotations only in structured details
- documented why the tool path intentionally does not use the slash-command idle guard
- updated PR body with dependency/notify-level context
- changed mode schema to direct/overlay literals
- inlined the trivial agent-review wrapper and simplified mode handling
- moved revdiff:launch emission into runReview
- made tool-opened result panels read-only to avoid returning stale state after panel actions

umputun (approved):
all round-1 points addressed in `96446ff`:

1. false-clean path at `revdiff.ts:519-535` now splits into three explicit branches (non-numeric exit, non-zero exit, missing output file), each with its own warning. Stricter than what I asked for.
2. version bumped to `0.2.0`.
3. tool result is short headline in `content`, full `ReviewState` only in `details`.
4. `ctx.isIdle()` guard path documented with a comment explaining why the tool runner serializes the interactive call.
5. PR body has the deprecation/notify-level context.

nits also all in: `mode` is a `Type.Union` of literals, `runAgentReview` wrapper inlined, `parseToolMode` removed, `revdiff:launch` emit moved into `runReview`.

copilot's openPanel concern handled by `readOnly: true` on the tool-opened panel: `ReviewPanel` gates `apply`/`rerun`/`clear` keys behind `!this.readOnly`, footer hint hides them, and `execute` no longer routes through `handleAction`.

lgtm, thx

PR #185, opened 2026-05-12
Configurable annotation marker

### LINKED ISSUE #184: Allow customising the annotation marker symbol
**Problem**
The `💬` emoji shown before inline annotations is hardcoded, so users who prefer a different indicator have no way to change it without forking.

**Solution**
Add an `--annotation-marker` flag (env: `REVDIFF_ANNOTATION_MARKER`, config: `annotation-marker`) that accepts a custom marker string. Default to `💬` to match current functionality.

```bash
revdiff --annotation-marker "▌"
# or in config:
# annotation-marker = ▌
```

The marker is used in two prefixes:
- Line-level: `<marker> ` before each annotation comment
- File-level: `<marker> file: ` before file-level annotations

Passing an empty string (`--annotation-marker ""`) should remove the marker entirely — annotations render with a bare space prefix. Multi-character markers should be supported (e.g. `>>>` is valid).

**Next steps**
I'm be happy to open a PR if you're happy with this direction and think it makes sense for the tool, and of course happy to discuss any details further!

jknlsn:
# Configurable annotation marker

Adds `--annotation-marker` (env: `REVDIFF_ANNOTATION_MARKER`, config: `annotation-marker`) to let users swap the `💬` prefix for something else. Defaults to the emoji so nothing changes for existing setups.

All five render sites now go through `annotPrefix()` / `annotFilePrefix()` helpers instead of hardcoding the emoji. Input width and wrap-continuation indent use `lipgloss.Width()` so everything stays aligned regardless of marker width.

The marker is render-only — structured output still uses `## path:N (T)` headers, so saved annotations stay portable between users with different markers. Flag lives on top-level `options` rather than `Colors` since it's not a colour value.

Empty string is a valid choice: `--annotation-marker ""` gives you a bare space prefix. The CLI default tag handles "not set", so empty only happens when you ask for it explicitly.

SKILL.md unchanged — this is a user-preference flag like `--theme` or `--wrap`, not something agents should auto-derive from context. Users set it in their config or env once.

Closes #184


umputun (changes requested):
few findings, biggest one is a width regression:

1. **input width drops 2 cols** at `app/ui/annotate.go:100,160`. Old constants `6`/`12` were `1 (cursor) + prefix_width + 2 (border margin)` per the original comments, the new `1 + lipgloss.Width(prefix)` drops the `+2`. With the default marker the input is now 4/10 instead of 6/12, two columns wider than before, can bleed into the right padding on narrow panes. Either restore the `+2` (`2 + lipgloss.Width(...)`) or document why the margin is no longer needed. Worth adding a regression test that pins absolute input width for the default marker plus at least one wide and one empty marker so this kind of drift surfaces in CI.

2. **testifylint will fail CI** at `app/config_test.go:427,454`. `assert.Equal(t, "", opts.AnnotationMarker)` should be `assert.Empty(t, opts.AnnotationMarker)`.

3. **gofmt struct-field alignment** at `app/main.go:206-213`, `app/ui/model_test.go:148-160`, `app/ui/annotate_test.go:510-520, 546-558`. Inserting `AnnotationMarker:` widened the longest key but the surrounding fields were not re-aligned. `make fmt` cleans it up.

4. **stale test comment** at `app/ui/annotate_test.go:1137`. `"(12-6=6)"` references the old hardcoded constants. The numerical result is still 6 under the new math (`10-4=6`), but the parenthetical no longer matches what is actually computed.

5. **weak assertion** in `TestModel_EmptyAnnotationMarkerExplicit`. `assert.Contains(t, rendered, " bare note", ...)` would pass even if a hardcoded `" 💬 "` fallback regressed since the substring would still appear after the prefix. Anchor with a regex or full-row equality on the prefix.

6. **per-render alloc** in `annotPrefix()` / `annotFilePrefix()`. Both concat a new string on every call from hot paths (`composeAnnotationRows`, `annotationContinuationIndent`, `annotationPrefixBody`, `renderAnnotationOrInput`, `renderFileAnnotationHeader`). Marker is immutable post-`NewModel`, would be cheaper to cache once on `modelConfigState` alongside `annotationMarker`.

umputun (commented):
all 6 round-1 items fixed, verified locally. Tests green, lint 0 issues.

two copilot inline comments from round 1 still worth a look:

1. **indent inference ambiguity** at `app/ui/diffview.go:701-710`. Real edge-case bug. `annotationContinuationIndent` uses `HasPrefix(firstLogicalLine, m.annotFilePrefix())` to pick indent width, but a line-level annotation whose body starts with `file: ` produces `💬 file: ...` and matches the file-prefix branch, so continuation rows get the 9-space indent instead of 3. Triggers on multi-line line-annotations only. Worse with empty marker since `annotFilePrefix` becomes `" file: "` and any body starting with `file: ` mis-matches. Fix is trivial since `composeAnnotationRows` already has `prefix` in scope: drop the `HasPrefix` switch and just do `indent := strings.Repeat(" ", lipgloss.Width(prefix))` inline, then either inline-delete `annotationContinuationIndent` or have it take `prefix` directly.

2. **no newline/control-char validation** on `--annotation-marker`. A `\n`/`\r`/`\t` in `REVDIFF_ANNOTATION_MARKER` (config file or env) would break layout silently. Nit, worst case is the user breaks their own rendering, but a `strings.ContainsAny(marker, "\n\r\t")` check at parse time with a clear error costs nothing.

re #5 (the empty-marker assertion): the `assert.NotContains(rendered, "💬")` you added catches a hardcoded emoji fallback regression, which was the actual worry. Fine as-is.

jknlsn:
Added fixes and tests for those copilot catches, thanks!

umputun (approved):
lgtm, thx

PR #189, opened 2026-05-12
Fix note about sandbox `excludedCommands`

plasticine:
This key moved from `permissions` to `sandbox` at some point


umputun (approved):
verified against the current Claude Code docs, `excludedCommands` is under `sandbox`. LGTM, thx

PR #194, opened 2026-05-13
ghostty skill: opt-in new-window mode (REVDIFF_GHOSTTY_MODE=window)

pavzagor:
## Problem

On Ghostty the launcher splits the focused window and calls `toggle_split_zoom`. The zoomed revdiff surface covers any existing pane in that window, so whatever the user had open there (often the shell or claude-code session that triggered the review) is hidden until revdiff exits.

## Change

Adds `REVDIFF_GHOSTTY_MODE` env var to `scripts/launch-revdiff.sh`:

- `split` (default) - current behavior, no change
- `window` - open revdiff in a new Ghostty window via AppleScript `new window with configuration`, so the original pane stays usable

Window mode uses `wait after command = true` (Ghostty's AppleScript-created windows don't honor `false` reliably, so we accept the brief "press any key" surface state). The launcher closes the window explicitly by its exact id once the sentinel fires.

The close osascript is `|| true`-guarded so a race where the window is gone before our close (user dismissed the press-any-key prompt manually) doesn't abort the launcher under `set -e` and lose the captured annotations.

## Safety

- default unchanged, opt-in only
- closes by exact window id captured at launch, never touches user's existing windows
- relies on `quit-after-last-window-closed=false` (Ghostty default), so closing the spawned window can't quit the app

## Demo

https://www.loom.com/share/102ac66f8b7b48f9991120e50890753a

## Test

`REVDIFF_GHOSTTY_MODE=window revdiff HEAD~1` in Ghostty 1.3.1 - new window opens with revdiff, annotations write to `--output`, window closes on quit. Default (`split`) unchanged.

### DIFF
--- a/.claude-plugin/skills/revdiff/scripts/launch-revdiff.sh
+++ b/.claude-plugin/skills/revdiff/scripts/launch-revdiff.sh
@@ -198,8 +198,13 @@ LAUNCHER
     exit 0
 fi
 
-# ghostty: split pane via AppleScript (macOS only, requires Ghostty 1.3.0+)
+# ghostty: AppleScript-driven overlay (macOS only, requires Ghostty 1.3.0+)
+# REVDIFF_GHOSTTY_MODE controls placement:
+#   split  (default) — split current window pane, zoom the new surface
+#   window           — open in a new Ghostty window so the original pane stays usable
 if [ "${TERM_PROGRAM:-}" = "ghostty" ] && command -v osascript >/dev/null 2>&1; then
+
+    GHOSTTY_MODE="${REVDIFF_GHOSTTY_MODE:-split}"
 
     SENTINEL=$(mktemp "$TMPBASE/revdiff-done-XXXXXX")
     rm -f "$SENTINEL"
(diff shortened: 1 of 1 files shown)


umputun:
the framing here doesn't match how this launcher actually works. Every overlay backend in `launch-revdiff.sh` is modal by design - tmux uses `display-popup -E`, zellij uses `--floating`, kitty uses `--type=overlay`, wezterm/cmux/iterm2 use splits, ghostty splits + zooms. All of them hide the caller pane while revdiff runs and return the user to it on exit. That's the design contract, not a ghostty-specific defect.

the use case itself isn't clear to me. Revdiff is modal by nature, that's the whole shape - take over the screen, review, dismiss, back to your shell. Making it "less modal" for one terminal isn't a fix, it's a different product. And if "i want my caller pane visible while reviewing" really is a need, it applies equally to kitty/tmux/zellij users, but nobody's asked for an escape hatch there. Adding `REVDIFF_GHOSTTY_MODE=window` would special-case ghostty for a personal preference and break the overlay consistency this launcher relies on.

closing as out-of-pattern.

Issue #187, opened 2026-05-12
symlink support for opencode plugin


elovelan:
Just wanted to float the idea, but I'm happy to skip it if it's too much of an edge case.

I'm using this on my fork since it's much easier to stay up-to-date with a simple `git pull`.

My thought was make this opt-in via a `--symlink` arg or similar to `plugins/opencode/setup.sh` ([example impl on my fork](https://github.com/elovelan/revdiff/blob/feat/opencode-setup-symlink/plugins/opencode/setup.sh#L13-L39))

umputun:
sure, this makes sense, opt-in `--symlink` with default copy preserved is the right shape. Pls send a PR from your fork branch.

a couple of asks:
- add a short note about the flag to `plugins/opencode/README.md`
- keep the default path byte-identical to current behavior

thx

umputun:
closing - approach didn't pan out, OpenCode's loader resolves symlinks before locating relative imports so the `.ts` plugins can't find their deps. See #198 / #199 for context and revert. `git pull && ./plugins/opencode/setup.sh` stays the supported update path.

PR #197, opened 2026-05-14
feat(opencode): add --symlink flag to setup.sh

elovelan:
## Problem

The OpenCode plugin setup script (`plugins/opencode/setup.sh`) currently copies all files into `~/.config/opencode/`. After every update to the revdiff repo, users must re-run the script to pick up changes. This is tedious for anyone developing or tracking the plugin closely.

## Solution

Add a `--symlink` flag to `setup.sh` that creates symlinks instead of copies. This lets updates propagate automatically after a simple `git pull` in the revdiff repo — no re-run needed.

## Changes

- `plugins/opencode/setup.sh`: Accept `--symlink` argument, use `ln -sf` instead of `cp` when the flag is set, rename internal helper from `copy_file` to `install_file`.
- `plugins/opencode/README.md`: Document the new flag and explain when to use it.

## Test Plan

- [x] Run `bash plugins/opencode/setup.sh --symlink` and verify symlinks are created in `~/.config/opencode/`.
- [x] Run `bash plugins/opencode/setup.sh` (without flag) and verify files are still copied.

### DIFF
--- a/plugins/opencode/README.md
+++ b/plugins/opencode/README.md
@@ -25,7 +25,15 @@
 bash setup.sh
 ```
 
-The script creates the target directories if needed, copies all files, marks the shell scripts as executable, and registers the plan-review plugin in `~/.config/opencode/opencode.json`. Or manually:
+The script creates the target directories if needed, copies all files, marks the shell scripts as executable, and registers the plan-review plugin in `~/.config/opencode/opencode.json`.
+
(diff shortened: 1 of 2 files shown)


umputun (approved):
lgtm. Both asks from #187 are in (opt-in flag, default path preserved, README note added).

one nit if you want to harmonize:

```sh
echo "Linked ${src#"$REPO_ROOT"/} as $dst"
```

vs the copy branch's `"Copied X -> $dst"`. Same shape `-> $dst` reads more consistent when a user runs both modes side by side. Not a blocker.

thx

PR #198, opened 2026-05-14
Revert "feat(opencode): add --symlink flag to setup.sh"

elovelan:
Reverting for the following reasons. Unfortunately, when I tested, everything "worked" because I had already run the setup script once without `--symlink` but I just re-ran it and realized it was only symlinking 3 files and not 5, which unfortunately also affects the origin logic path as well.

- Bash returns the return value of the last statement. Because I moved the `echo` into the `if` statement, it was failing on the first call to the function without the `exec` arg due to `[[[ "$make_exec" == "exec" ]] && chmod +x "$dst"` returning false.
 
  Ideally, this would be fine and just report a false set of errors at the end, but `((errors++))` also returns a non-zero exit code (interestingly `((++errors))` does not), so the entire script was failing at that point.
- Once I fixed the issue above, I ran into an issue where OpenCode was no longer working because their loader resolves symlinks and it couldn't find the necessary libraries relative to the repo, so this whole thing is a non-starter.

In an ideal scenario, this gets shipped as a "real" plugin, but I think automating `git pull && ./plugins/opencode/setup.sh` is fine for now.

Again, my sincere apologies for my oversight in testing, I should have been more thorough.

### DIFF
--- a/plugins/opencode/README.md
+++ b/plugins/opencode/README.md
@@ -25,15 +25,7 @@
 bash setup.sh
 ```
 
-The script creates the target directories if needed, copies all files, marks the shell scripts as executable, and registers the plan-review plugin in `~/.config/opencode/opencode.json`.
-
-Use the `--symlink` flag to create symlinks instead of copies. This makes updating as simple as a `git pull` inside this repo — no need to re-run the script after every change:
-
-```sh
-bash setup.sh --symlink
-```
-
-Or manually:
+The script creates the target directories if needed, copies all files, marks the shell scripts as executable, and registers the plan-review plugin in `~/.config/opencode/opencode.json`. Or manually:
 
 ```sh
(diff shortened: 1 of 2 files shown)


umputun:
both findings are right, and the bash one is on me too - I noted the ordering swap in review but called it "no real-world effect", missing that the function's last statement now returns the `chmod` short-circuit instead of the always-succeeding `echo`. Combined with `((errors++))` returning status 1 on `errors=0`, `set -e` aborts the script. Should have caught that.

opened #199 to revert. The OpenCode loader behavior is a real blocker for the whole approach, so closing the symlink idea here. `git pull && ./plugins/opencode/setup.sh` stays the supported update path.

no apology needed - thx for digging in and reporting back.

elovelan:
I'll be tracking anomalyco/opencode#23553, though I'm not sure that it really makes that much of a difference to move to `git pull` from `git pull && ./plugins/opencode/setup.sh` 😆

Issue #120, opened 2026-04-18
Issue: duplicated terminal support code


nborodikhin:
When Claude Code executes a child process it does so without allocating a PTY which `revdiff`, being an interactive program, requires.

There are workarounds for this issue that use a custom per-terminal approach to create a new pane or overlay and launch `revdiff` in a new environment.  The issue is that there are many places that start `revdiff` and per-terminal workarounds have to be updated manually every time a new terminal support is implemented:

- revdiff diff review at `.claude-plugin/skills/revdiff/scripts/launch-revdiff.sh` |
- revdiff codex skill at `plugins/codex/skills/revdiff/scripts/launch-revdiff.sh`  |
- revdiff plan review at `plugins/revdiff-planning/scripts/launch-plan-review.sh`  |
- external projects, such as `cc-thingz` [planning plugin](https://github.com/umputun/cc-thingz#planning)

umputun:
thx, but this is a deliberate design decision, not a bug.

Claude Code skills are supposed to be self-contained — unifying scripts across skills doesn't work reliably, at minimum you hit permission prompts on every invocation, and it also raises the question of how a skill with a "common" external dependency gets distributed and updated. Same story for the codex skill and the revdiff-planning plugin — each needs to stand on its own.

In practice what we have works well enough. Agents are aware of the duplication and CLAUDE.md has explicit instructions for keeping the copies in sync when terminal support changes. Not ideal, but the alternatives are worse.

Also per CONTRIBUTING.md, general ideas and brainstorming belong in [Discussions](https://github.com/umputun/revdiff/discussions), not Issues. Moving this to Discussions — pls continue there if you want to explore a concrete shape.

PR #200, opened 2026-05-14
feat(app): add shell completion script generation

kunzaatko:
Add a --completions flag that prints bash, zsh, or fish completion
scripts and wires shell-name completion into the flag itself.
Includes tests for template output and identifier sanitisation.

There was a necessary workaround because the CLI does not implement subcommands which are what is completed when there
is nothing after the command so the completion scripts for the shells add a "--" to the completion command and then
fetches the completions.

image

image

## Release Notes

* **New Features**
  * Added shell auto-completion support for bash, zsh, and fish shells through a new `--completions` flag.
  * Generate customized completion scripts for your preferred shell to enable intelligent command-line auto-completion.
  * Completion scripts provide smart command suggestions, streamline command discovery, and reduce manual typing for enhanced CLI productivity.

[Review Change Stack](https://app.coderabbit.ai/change-stack/umputun/revdiff/pull/200?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/app/completions.go
+++ b/app/completions.go
@@ -1,0 +1,84 @@
+package main
+
+import (
+	"fmt"
+	"io"
+	"path/filepath"
+	"strings"
+	"unicode"
+)
+
+func printCompletionScript(w io.Writer, shell, bin string) error {
+	cmd := filepath.Base(bin)
+	name := sanitizeShellIdent(cmd)
+	switch shell {
+	case "bash":
+		fmt.Fprintf(w, bashCompletionTemplate, name, bin, name, bin)
+	case "zsh":
+		fmt.Fprintf(w, zshCompletionTemplate, bin, name, bin, name, bin)
+	case "fish":
+		fmt.Fprintf(w, fishCompletionTemplate, name, bin, cmd, name)
+	}
+	return nil
+}
+
+// sanitizeShellIdent turns a binary name into a valid shell identifier.
+func sanitizeShellIdent(s string) string {
+	var b strings.Builder
+	for i, r := range s {
+		if unicode.IsLetter(r) || r == '_' || (i > 0 && unicode.IsDigit(r)) {
+			b.WriteRune(r)
+		} else {
+			b.WriteRune('_')
+		}
+	}
+	return b.String()
+}
+
+// bashCompletionTemplate aligns with the official go-flags bash completion
+// example, with one addition: when the current token is empty we inject "--"
+// so go-flags completes option names rather than falling through to positional
+// completion (our positional args are plain strings with no Completer).
+var bashCompletionTemplate = `_%s() {
+    local args=("${COMP_WORDS[@]:1:$COMP_CWORD}")
+    if [ ${#args[@]} -eq 0 ] || ([ ${#args[@]} -eq 1 ] && [ -z "${args[0]}" ]); then
+        args=("--")
+    fi
+    local IFS=$'\n'
+    COMPREPLY=($(GO_FLAGS_COMPLETION=1 %s "${args[@]}"))
+    return 0
+}
+complete -F _%s %s
+`
+
+// zshCompletionTemplate aligns with the open PR #414 for go-flags zsh
+// completion, with the same empty-token "--" injection.
+var zshCompletionTemplate = `#compdef %s
+_%s() {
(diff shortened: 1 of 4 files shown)


umputun (changes requested):
the main issue is strategic, not implementation. The canonical shell-completion pattern for similar projects is the one paskal shipped across ~17 of them (ralphex #120, reproxy #244, mpt #33, spot #337, stash #69, sys-agent #53, secrets #105, tg-spam #368, and others). The closest cousin is ralphex, which uses go-flags' built-in `GO_FLAGS_COMPLETION=1` mechanism.

the pattern there:

- three small static scripts in `completions/<name>.{bash,zsh,fish}` (~25 lines total)
- each invokes the binary with `GO_FLAGS_COMPLETION=1`, no templating in Go code
- `.goreleaser.yml` ships them in release archives, deb/rpm, and the homebrew formula (auto-installed)
- one conditional in `main.go` to suppress the startup banner when `GO_FLAGS_COMPLETION` is set
- total ~41 lines, 5 files

this PR reinvents that with 5x the code, hand-rolled templates inside Go, and the resulting bugs Copilot flagged (unquoted `os.Args[0]`, discarded `Fprintf` errors, zsh passing wrong word slice, silent no-op on unsupported shell) don't exist in the static-file approach. The `--` workaround isn't needed either - go-flags handles bare TAB natively.

would prefer to align this PR with paskal's pattern (ralphex #120 is the reference) or close it in favor of someone porting the canonical approach.

kunzaatko:
@umputun I will make the changes next week. I genuinely think that the "hack" with the "--" is necessary though. I tested it out without it and the completions do not show up. Can you review that please? I partially agree that the way that you suggest is better than the one that I suggested.

umputun:
verified the claim. You're right that without injection, `GO_FLAGS_COMPLETION=1 revdiff ""` returns empty - go-flags routes empty input to positional completion (`completion.go:279`), and `Refs.Base` is a plain `string` with no `Completer`, so the list comes back empty. Ralphex has the same shape (`PlanFile string` positional, no completer).

so the workaround does what you say it does. My read on whether it's *needed* is different though: the canonical pattern handles bare TAB by falling back to **file completion**, not by forcing the flag list. Look at ralphex's three scripts:

`completions/ralphex.bash`:
```bash
complete -o default -F _ralphex ralphex
```
the `-o default` is the trick. When COMPREPLY is empty, bash falls back to its default file completion.

`completions/ralphex.zsh`:
```zsh
if (( ${#lines} )); then
    _describe 'command' descr
else
    _files
fi
```
explicit `_files` fallback.

`completions/ralphex.fish` - omits `-f` so fish keeps default file completion.

result: on bare TAB, ralphex shows files (useful when the positional is a path or ref). On `-<TAB>` it shows flags. That matches the convention most CLI tools follow - bare TAB completes the positional, `-<TAB>` shows options.

couple other things worth picking up while porting:

- ralphex uses `GO_FLAGS_COMPLETION=verbose` (not `=1`) for zsh/fish - emits flag descriptions alongside names, much more useful in completion popups
- the scripts are tiny static files in `completions/`, no Go templating, no sanitization, no `--completions` flag, no `os.Args[0]` handling, and `.goreleaser.yml` ships them via release archives, deb/rpm, and homebrew formula

this is my read based on reading the ralphex scripts and testing locally, but paskal authored the pattern across ~17 projects so he's the authority on the design choices. /cc @paskal for a sanity check on whether bare-TAB-shows-files vs bare-TAB-shows-flags is intentional UX or just incidental.

umputun:
going with the canonical go-flags completion pattern from the review (ralphex #120) instead of the hand-rolled templates. Put it up as #226 - three static scripts, no Go code, and it avoids the os.Args[0] quoting, discarded write errors, and bare-TAB handling that came up here.

your testing of the `--` workaround was useful - it confirmed go-flags returns empty on bare TAB, which #226 handles with file-completion fallback (`-o default` / `_files`) rather than forcing the flag list. Closing this in favor of #226, but the feature was your idea, so credit for that.

umputun:
#226 is merged to master. Shell completions for bash, zsh, and fish now ship via the release packages (deb/rpm/brew auto-install them).

PR #205, opened 2026-05-21
feat: Add H/M/L screen-position motions to vim-motion preset

andykog:
Adds `H`, `M`, `L` screen-position cursor motions to the vim-motion preset, matching standard vim behavior:

- `H` — move cursor to the top of the visible screen
- `M` — move cursor to the middle of the visible screen
- `L` — move cursor to the bottom of the visible screen

Supports count prefixes: `5H` jumps to the 5th line from the top, `3L` to the 3rd line from the bottom.

These are diff-pane only and require `--vim-motion` to be enabled. In the tree pane they fall through to normal bindings.

Note that **`L` overrides `toggle_line_numbers`** in vim-motion mode (rebindable via keybindings file).


umputun (changes requested):
the H/M/L feature makes sense, fills the obvious gap next to gg/G/zz/zt/zb. Few things to sort before merge.

blocking:

1. lint fails. `diffnav.go:352` `if row < m.layout.viewport.YOffset { row = ... }` trips the `modernize` linter, use `row = max(row, m.layout.viewport.YOffset)`. (CI doesn't run on fork PRs in this repo, so it wasn't caught here.)

2. no tests for the new motions. `vimmotion_test.go` and `diffnav_test.go` both exist but neither gets an H/M/L case, only the help-overlay count assertion in `handlers_test.go` changed. Need coverage for `5H`/`3L` placement, `M` on odd/even heights, cursor on a divider row, empty diff, count-cleared-after-use. The existing `G` motion tests are a good template.

3. the three motions discard the `onAnnotation` bool from `visualRowToDiffLine`: `idx, _ := ...` then `m.annot.cursorOnAnnotation = false` (diffnav.go:319/335/354). When H/M/L target a visual row that is an annotation sub-line, the cursor ends up on the parent diff line instead, the highlight paints the wrong row and a later annotation edit/delete acts on the wrong target. Should be `idx, onAnn := ...; m.annot.cursorOnAnnotation = onAnn`.

4. `nudgeCursorOffDivider` searches backward-first with no viewport bound. `H` targets `YOffset` exactly, which is often a `⋯ N lines ⋯` divider, so the backward search can pick a line above the viewport and `syncViewportToCursor` then scrolls the page. For a screen-position motion the nudge should stay inside the visible row range, pls constrain it (prefer the in-viewport direction).

non-blocking:

5. `nudgeCursorOffDivider` re-implements the backward-then-forward divider scan already inlined in `jumpToLineN` (diffnav.go ~409-425). Worth having `jumpToLineN` call the helper so there is one copy. The order differs too, `jumpToLineN` does adjustCursorIfHidden then nudge, the new methods nudge then adjust, pls reconcile.

6. the `n := m.vim.count; m.vim.count = 0; m.vim.hint = ""` block is repeated across the G/H/M/L cases. An extracted `consumeVimCount()` would keep it in one place.

`L` shadowing `toggle_line_numbers` in vim-motion mode is fine, it is an opt-in preset, but pls add a one-line note to the vim-motion section in README so it is discoverable outside the PR description.

andykog:
Thanks @umputun, I was going to fix your comments but you already did it 🙌

PR #210, opened 2026-05-23
fix(zellij): route popups to originating tab via pane tracking

danruto:
## Summary

Fixes Zellij popups (revdiff review + plan review) appearing in the *currently focused* tab rather than the tab where the user invoked the command. If the user switched tabs between firing the action and the popup spawning, it would land out of
sight — looking like a silent failure.

  ## Problem

`zellij run --floating` always targets the focused tab at spawn time. When Claude Code (or a hook) takes a few seconds to fire, the user often tabs away — and the popup opens in whichever tab they're now looking at, or worse, in a tab they've
since left, where it sits invisible until they wander back.

  ## Fix

Resolve the originating pane's tab up front and route the popup there explicitly:

1. Read `ZELLIJ_PANE_ID` (exported by zellij 0.40+ into the calling shell).
2. Query `zellij action list-panes --json` and use `jq` to look up the matching `tab_id`, filtering on `is_plugin=false`
(plugin panes share id numbers with terminal panes).
3. Spawn the popup with `zellij action new-pane --floating --tab-id <id>` so it lands in the originating tab regardless of current focus.
4. Fall back to the original `zellij run --floating` path when `ZELLIJ_PANE_ID` or `jq` is missing — no behavior change for older zellij or systems without jq.

### DIFF
--- a/.claude-plugin/skills/revdiff/scripts/launch-revdiff.sh
+++ b/.claude-plugin/skills/revdiff/scripts/launch-revdiff.sh
@@ -136,10 +136,33 @@ $(write_rc_cmd "$SENTINEL")
 LAUNCHER
     chmod +x "$LAUNCH_SCRIPT"
 
-    zellij run --floating --close-on-exit \
-        --width "$POPUP_W" --height "$POPUP_H" \
-        --name "$OVERLAY_TITLE" --cwd "$CWD" \
-        -- "$LAUNCH_SCRIPT" >/dev/null 2>&1
+    # zellij run targets the focused tab; route into our originating tab via
+    # `action new-pane --tab-id` so the popup follows the caller's pane even
+    # when the user has switched tabs. needs ZELLIJ_PANE_ID (zellij 0.40+) + jq;
+    # falls back to the plain `zellij run` path otherwise.
+    #
+    # `list-panes --json` reports `.id` as a bare integer and plugin panes
+    # share id numbers with terminal panes, so filter on is_plugin=false.
+    ZELLIJ_TAB_ID=""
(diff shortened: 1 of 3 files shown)


umputun (commented):
couple issues before merge:

1. I don't think this needs the `jq`/`list-panes` path yet. `zellij run --help` already has `--near-current-pane` and `--tab-id`, so pls try the smaller `zellij run --near-current-pane` fix first, or show why it doesn't solve the tab-routing problem.

2. current fallback can fail under `set -euo pipefail`. If `zellij action list-panes` or `jq` fails, the command substitution exits the launcher instead of falling back to `zellij run`.

3. `list-panes --json` doesn't include tab fields unless `--tab` is passed, at least on zellij 0.44.3. Without that, `.tab_id` can be missing and `jq -r` can return `null`, which passes `[ -n "$ZELLIJ_TAB_ID" ]` and runs `--tab-id null`.

4. if the targeted `new-pane --tab-id` call fails after a tab id was found, the old `zellij run` path is not tried.

I'd prefer this to stay on `zellij run` if possible, with the old behavior kept as an actual fallback.

danruto:
Thanks for the feedback. It makes the script update alot simpler.

I've updated the script to use `zellij run`. We need need jq just to fetch `tab-id` otherwise the UX isn't ideal e.g. it will either take over whatever you were doing or spawn the popup in the background.

With `--tabs` I didn't need it on `0.44.3` but there is no hard in adding it in.

umputun (approved):
lgtm.

minor: PR body still mentions `zellij action new-pane`, but the current code uses `zellij run --tab-id`.

PR #216, opened 2026-05-27
feat: parse multi-file git diffs piped to stdin

asek-ll:
Several version control systems (and assorted tooling) can emit
git-compatible unified diffs. Detecting and parsing a git-style diff on
stdin lets revdiff review the output of any such tool without a dedicated
renderer per VCS — supporting the git diff format once is cheaper to
maintain than wiring up a new VCS integration every time one appears.

Previously stdin was always treated as raw, context-only text. It is now
sniffed for a unified-diff signature, split into per-file sections, and
parsed into real diff lines, falling back to the raw-text reader when
detection or parsing fails.

### DIFF
--- a/.claude-plugin/skills/revdiff/SKILL.md
+++ b/.claude-plugin/skills/revdiff/SKILL.md
@@ -42,6 +42,25 @@ The script resolves the history dir from `$REVDIFF_HISTORY_DIR` (default `~/.con
 
 When the user asks to open an in-session review in revdiff (the conversation already contains review comments produced earlier in the session), write those comments to a temp file (e.g. `/tmp/revdiff-review-XXXXXX.md`) using the format documented in `references/usage.md` ("Output Format" section), then run the normal launcher flow (Step 1 ref detection, Step 2 invocation) with `--annotations=<temp-path>` appended. Step 3 onward handles the curated annotations as usual.
 
+## Reviewing a Diff That Lives Outside the Working Tree
+
+Some review targets are not the current repo state: a GitHub PR diff, a patch file on disk, or `git format-patch -1 --stdout` output. Pipe the unified diff into `revdiff --stdin` and the input is parsed as a real multi-file diff (one tree entry per file, hunk navigation, per-file annotations) instead of a context-only buffer. revdiff auto-detects the unified-diff signature; on a malformed patch the input falls back silently to raw-text mode.
+
+Use this instead of the normal launcher flow when:
+- the user asks to "review PR #N", "review this patch", "review `gh pr diff` output", or supplies a patch URL/path
+- the diff describes commits that are not checked out locally (e.g. someone else's branch on a remote-only PR)
+- the user pastes a unified diff and asks for a review of *that diff*, not the working tree
+
+Example invocations (route through the same launcher resolver as the normal flow):
+
+```bash
+gh pr diff 123 | "$("${CLAUDE_SKILL_DIR}/scripts/resolve-launcher.sh" launch-revdiff.sh "${CLAUDE_PLUGIN_DATA}")" --stdin
+git format-patch -1 --stdout | "$("${CLAUDE_SKILL_DIR}/scripts/resolve-launcher.sh" launch-revdiff.sh "${CLAUDE_PLUGIN_DATA}")" --stdin
+cat /tmp/feature.patch | "$("${CLAUDE_SKILL_DIR}/scripts/resolve-launcher.sh" launch-revdiff.sh "${CLAUDE_PLUGIN_DATA}")" --stdin
+```
+
+`--stdin` is mutually exclusive with refs, `--staged`, `--only`, `--all-files`, `--include`, `--exclude`, and `--annotations`, so do not combine with the Step 1 ref detection — go directly to Step 3 once the launcher returns. Annotations come back keyed by the real file paths from the diff (not by `--stdin-name`).
+
 ## How It Works
 
 1. Launch revdiff in a terminal overlay (tmux popup, Zellij floating pane, kitty overlay, wezterm/Kaku split-pane, cmux split, ghostty split+zoom, iTerm2 split pane, or Emacs vterm frame)
--- a/.claude-plugin/skills/revdiff/references/usage.md
+++ b/.claude-plugin/skills/revdiff/references/usage.md
@@ -99,11 +99,17 @@ revdiff --compare-old=a.txt --compare-new=b.txt
 
 ## Scratch-Buffer Review
 
-Use `--stdin` to review arbitrary piped or redirected text as one synthetic file. All lines are treated as context, so single-file mode, inline annotations, file-level notes, search, wrap, collapsed mode, and structured output all work unchanged.
+Use `--stdin` to review arbitrary piped or redirected text. revdiff sniffs the input for a git unified-diff signature: when a line beginning with `diff --git a/` is found near the start, the input is parsed as a real multi-file diff (one tree entry per file, with `+`/`-` markers, hunk navigation, word-diff, compact mode, per-file annotations). Otherwise the input is shown as a single context-only buffer — single-file mode, inline annotations, file-level notes, search, wrap, collapsed mode, and structured output all work unchanged.
 
 - `--stdin` is explicit and requires piped or redirected input
-- `--stdin-name` sets the synthetic filename used in annotations and syntax highlighting
+- `--stdin-name` sets the synthetic filename used by the context-only buffer (ignored in multi-file diff mode, where real paths are shown)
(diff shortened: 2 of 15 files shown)


umputun (commented):
interesting direction and the parse-then-fallback shape is right. Couple of things worth changing before this merges.

**blocking**

1. **silent skip of failed sections** at `app/diff/stdin.go:81-97`. If one section out of N fails `parseUnifiedDiff`, it's `continue`d with an `fmt.Fprintf(os.Stderr, "warning: ...")` and the renderer is still returned as success. The stderr warning is invisible under bubbletea's alt screen. A user piping a 20-file patch where one file's hunks malform sees 19 files in the tree with no in-TUI hint the 20th is gone. That's a review-integrity bug. Pls make any per-section parse error fail the whole `NewMultiFileStdinReader` call so the caller falls back to raw-text mode for the entire input.

2. **no size cap on stdin** at `app/stdin.go:85`. `io.ReadAll(stdin)` is uncapped, then `string(content)` is taken 2-3 times (lines 91, 93, 99) so the buffer is held ~3x briefly. Pre-PR was uncapped too, but only the final `[]DiffLine` survived. `cat /dev/zero | revdiff --stdin` is worse now. Project precedent is `--description-file` using `io.LimitReader` with a clear error on overflow. Pls cap (64-256 MB is fine) and also do `s := string(content)` once and reuse it.

3. **sniffer is too loose** at `app/diff/multiparse.go:19-38`. Godoc says "starts with `diff --git a/`" but the code is `strings.Contains` over the first 4 KB, and the `@@ -` secondary fires on any text containing a quoted hunk header. A markdown file or chat log mentioning `diff --git a/x b/x` literally in a code block gets reclassified and surrounding prose is dropped from the rendering. Pls anchor to line-start (`^diff --git a/` with newline-boundary, or strings split + prefix check on the first non-blank line) and either drop the `@@ -` fallback or anchor it the same way. That also makes the godoc accurate.

4. **docs missing**. Per the project rule, behavior change to an existing flag is the same documentation obligation as a new flag. Pls add:
   - `README.md` `--stdin` section (currently says "All lines are shown as context" which is no longer accurate)
   - `site/docs.html` matching paragraphs
   - `.claude-plugin/skills/revdiff/references/usage.md`
   - `.claude-plugin/skills/revdiff/SKILL.md` (when AI agents should pipe `gh pr diff` / `git format-patch -1 --stdout` for multi-file review)
   - `plugins/codex/skills/revdiff/SKILL.md` and `plugins/pi/skills/revdiff/SKILL.md` (kept byte-identical with the Claude copy)
   - `docs/ARCHITECTURE.md` (Renderer table near line 386, add `MultiFileStdinReader` row)

   No plugin version bump on this branch though, the feature is in the binary and the version-defer rule applies.

**non-blocking**

5. `fmt.Fprintf(os.Stderr, ...)` from inside `app/diff/` under a TUI doesn't fit the project's `log.Printf` convention. Once #1 above is in, the whole warning probably goes away.

6. `app/diff/stdin.go` now mixes `StdinReader` and `MultiFileStdinReader` in one file but tests are already split (`stdin_test.go` vs `multiparse_test.go`). Worth moving `MultiFileStdinReader` to a `multistdin.go` to keep the 1:1 source-to-test convention.

7. `NewStdinReaderFromString` (`stdin.go:31`) has one caller in `app/stdin.go:99` that hands it a string. It can be inlined as `NewStdinReaderFromReader(name, strings.NewReader(content))` and the constructor dropped. Same shape for `IsUnifiedDiff` (one caller, can be unexported `isUnifiedDiff`).

8. the error from `NewMultiFileStdinReader` is dropped at `app/stdin.go:93` (`if multi, mErr := ...; mErr == nil`). With #1 above in place this becomes load-bearing, pls don't silence it, at minimum `log.Printf` it.

9. nit: `splitMultiFileDiff` resolves path/status via `parseFileHeader`, then `parseUnifiedDiff` walks the same header text per section. Not a bug, but the package now has two recognizers for `diff --git` (regex in `multiparse.go:45` vs inline skip in `diff.go:668`). Folding the section split into `parseUnifiedDiff` so it emits `(path, status, []DiffLine)` tuples directly would dedupe, but that's a wider refactor than this PR needs, mentioning it for awareness.

10. nit: `sec.path` is attacker-controlled and gets `fmt.Fprintf`'d to stderr unsanitized. Once #1 is in the warning's gone, but if any other code prints diff-header paths later, route through `diff.SanitizeCommitText` like the rest of the project.

**tests worth adding**

- markdown / chat-log containing `@@ -1,5 +1,5 @@` or `diff --git a/x b/x` in a fenced block (locks in the tightened sniffer from #3)
- fallback end-to-end: sniff returns true, parse fails on all sections, caller falls back to raw text (locks in #1's behavior)
- malformed `diff --git` headers (truncated, missing `b/` side) that hit the empty-path-discard branch in `splitMultiFileDiff`
- renames where the `rename to` path contains spaces
- large input (would have caught #2)

tests/lint/race all green locally, no scope creep, code is idiomatic Go. Main concerns are the four blocking items above, addressing the non-blocking ones too would be appreciated.

umputun (commented):
all four blocking items are addressed:

1. parse-failure now fails the whole `NewMultiFileStdinReader`, so the caller falls back to raw text (`multistdin.go:48-53`)
2. 64 MiB cap via `io.LimitReader(r, cap+1)`, enforced before the single `string(data)` (`readStdinCapped` in `app/stdin.go`)
3. `isUnifiedDiff` is line-anchored now (needs a line starting with `diff --git a/`), `@@ -` fallback dropped, prose-mention test added
4. docs synced across all six surfaces

non-blocking 5/6/7/8/10 are in too.

**one blocking thing, #1 isn't fully closed.** The parse-failure path is fixed, but two sibling cases still produce a degraded tree instead of raw-text fallback, which is the same review-integrity hazard:

1. **split-layer drop** (`multiparse.go:673-683`): `splitMultiFileDiff`'s flush does `if path == "" { return }`, so a `diff --git ` boundary whose header yields no path is dropped silently. Valid section + a malformed second section means the reader succeeds with the second file missing from the tree, no fallback.
2. **empty-content section** (`multistdin.go`): a section that sniffs and splits but where `parseUnifiedDiff` returns zero lines (e.g. a doc whose first line is `diff --git a/foo b/foo` with no hunks) is accepted as an empty file, so the real stdin text is hidden behind an empty tree entry instead of being shown as raw.

both reduce to one fix: a section that produces nothing renderable should be treated as a parse failure so the whole input falls back to raw text.

pls add two regression tests as well - valid + malformed section falling back to raw, and a line-start `diff --git` with no hunk falling back to raw.

**non-blocking:**

- `ChangedFiles` (`multistdin.go`) can emit duplicate tree rows - `order` is a `[]string` recording every path, so two sections resolving to the same path show up twice. Same crafted-diff family as above, low severity. Dedupe when appending to `order`.
- `rename to` (`multiparse.go:141-145`) bypasses `cleanPath`, while every other branch routes through it, so a git-quoted `rename to "..."` keeps its quotes. Route it through `cleanPath` too.
- `cleanPath` (`multiparse.go:709`) double-strips a top-level dir literally named `a`/`b`, so `b/weird.go` resolves to `weird.go`. Rare, but the triple `TrimPrefix` is the cause.
- `stdin_test.go` is new but only tests `MultiFileStdinReader` (which lives in `multistdin.go`), while `stdin.go`'s type is tested elsewhere. Rename to `multistdin_test.go` per the one-test-file-per-source rule.
- `selectStdinRenderer` godoc (`app/stdin.go`) says the WARN logs "despite the alt-screen swallowing stderr", but it runs before `p.Run()` so stderr is visible. The logging is fine, the rationale in the comment is wrong, drop that clause.
- `app/diff/stdin.go` wraps `import "io"` in parens for no functional reason, looks like unrelated churn, revert it.
- the 4096-byte sniff window misclassifies `git format-patch` output when the commit body pushes the first `diff --git` past 4 KiB. Fine tradeoff for the documented `git diff | revdiff --stdin` case, but it's neither in the godoc nor covered by a test.

the integrity fallback is the only real gate here, the non-blocking items can ride along in the same push or a follow-up.

umputun (approved):
all addressed, thx.

the blocking fallback-integrity hole is fully closed now. Empty-path sections fail at the split layer (`multiparse.go` flush returns an error), and hunkless-but-non-structural sections fail in `NewMultiFileStdinReader` via `hasStructuralChange` - both route the whole input back to raw text. The structural-marker carve-out keeps rename-only / mode-only / new-empty sections valid, which is the right call. Regression tests cover both the valid+malformed section and the line-start `diff --git` with no hunk.

the 7 non-blocking nits are all in too: order dedupe, `rename to` through `cleanPath`, single-strip `cleanPath`, test-file rename, corrected `selectStdinRenderer` godoc, import revert, and the 4 KiB sniff-window doc + test.

tests/lint/race green locally, lgtm.

PR #224, opened 2026-06-06
feat: scroll the diff view with J/K hotkeys

steve-mackinnon:
This mimics behavior in lazygit: you can quickly scroll through the diff without changing focus to the diff view. I've found this is a nice improvement as I can quickly change files and scroll through their diffs with fewer keystrokes. I opted to scroll by 2 lines at a time for parity with lazygit.

- Adds action items to the help page
- Adds tests to verify the behavior with both tree and diff view focused


umputun (changes requested):
tests/lint/race all green, CI green, no correctness issues. Copilot found nothing either. Three asks before merge:

1. **reuse the wheel step.** `diffScrollStep = 2` now sits next to `WheelStep = 3` (mouse wheel) and the shift+wheel half-page, so there are three different small-scroll steps in the same space. Drop `diffScrollStep` and pass the existing `wheelStep` into `scrollDiffViewportLine` so J/K matches the wheel. Lazygit uses 2, but staying consistent with revdiff's own wheel step is better here.

2. **docs need updating in this PR.** The in-app help auto-generates from `defaultDescriptions()` so the overlay is covered, but the hand-maintained docs are stale. New bindings/actions need to go into:
   - keybindings tables: `README.md`, `site/docs.html`, `.claude-plugin/skills/revdiff/references/usage.md`, `plugins/codex/skills/revdiff/references/usage.md`
   - actions lists: `README.md`, `site/docs.html`, `.claude-plugin/skills/revdiff/references/config.md`, `plugins/codex/skills/revdiff/references/config.md`

   the two plugin copies must stay byte-identical to each other.

3. **assert the cursor pin.** `TestModel_JKScrollsDiffViewport` checks `YOffset`, focus, and tree selection but never `m.nav.diffCursor`. The pin is the reason `scrollDiffViewportLine` exists over the plain `scrollDiffViewportBy`, and the test would still pass if the pin were dropped. Add a `diffCursor` assertion after the J press.

steve-mackinnon:
@umputun thanks for the quick feedback. Just made the requested changes - ready for another look.

umputun (approved):
all three asks addressed: wheel-step reuse (`diffScrollStep` dropped, J/K uses `wheelStep`), docs across all six locations (the two `config.md` copies byte-identical), and the `diffCursor` pin assertion after the J press.

PR #229, opened 2026-06-14
feat(plugins): render revdiff in a tmux window under agent-deck

paskal:
## What

Under [agent-deck](https://github.com/asheshgoplani/agent-deck), revdiff's review overlay never appears if I'm in another session when revdiff is spawned.

agent-deck *does* surface real tmux **windows**. This PR adds a small backend that, when agent-deck is detected, runs revdiff in a background tmux window (named after the content under review) instead of a popup. It shows up in the agent-deck session tree for the user to switch to, and never pops over whatever other session they're working in.

## How

- New `agentdeck-window.sh`, **sourced** by `launch-revdiff.sh` — it reuses the existing `REVDIFF_CMD` / `OUTPUT_FILE` / `sq` / `write_rc_cmd` / `read_rc` / `print_output_and_exit` plumbing and the sentinel/rc protocol, so the exit-code-on-annotations behaviour is preserved.
- `launch-revdiff.sh` gains only a 4-line guarded `source` hook before the tmux popup path. Every non-agent-deck tmux returns immediately and uses the popup **unchanged** — no behaviour change and no added cost for existing users.
- Detection: `AGENTDECK_INSTANCE_ID` (also mirrored into the tmux session env) with the `agentdeck_*` session-name prefix as a fallback.
- Window created with `new-window -d` so it never steals the active window.
- Applied to both the main (`.claude-plugin/…`) and codex (`plugins/codex/…`) plugin script sets.

## Opt in / out

- `REVDIFF_TMUX_WINDOW=1` — force window mode (any tmux)
- `REVDIFF_TMUX_WINDOW=0` — force the popup path
- unset — auto: window mode only when agent-deck is detected

## Testing

- `bash -n` + `shellcheck -x` clean on all four changed files.
- Verified end-to-end inside agent-deck: the review window opens in the background (active window not stolen), annotations are captured on close via the existing sentinel/rc flow, and exit-code-on-annotations is preserved.

How it looks like:

Screenshot 2026-06-15 at 00 02 53

### DIFF
--- a/.claude-plugin/skills/revdiff/scripts/agentdeck-window.sh
+++ b/.claude-plugin/skills/revdiff/scripts/agentdeck-window.sh
@@ -1,0 +1,58 @@
+# shellcheck shell=bash
+# revdiff — agent-deck backend. SOURCED by launch-revdiff.sh (never run standalone).
+#
+# Why this exists: agent-deck (https://github.com/asheshgoplani/agent-deck) renders each of its
+# sessions through a tmux *control-mode* client, which does NOT render `tmux display-popup` at
+# all — the popup is invisible to the user and `display-popup -E` blocks forever (the agent
+# waits on a review the human can never see). agent-deck DOES surface real tmux windows, so
+# under agent-deck revdiff runs in a background tmux window named after the content under
+# review. It appears in the agent-deck session tree for the user to switch to when ready and
+# never pops over whatever other session they are working in.
+#
+# Activation:
+#   REVDIFF_TMUX_WINDOW=1   force window mode (any tmux)
+#   REVDIFF_TMUX_WINDOW=0   force the popup path (skip this backend)
+#   unset                   auto: window mode only when agent-deck is detected
+#
+# Reuses from the caller (launch-revdiff.sh): TMPBASE, OUTPUT_FILE, REVDIFF_CMD, CWD, DIR_NAME,
+# TITLE_REF, and the helpers sq() / write_rc_cmd() / read_rc() / print_output_and_exit().
+# The caller guarantees $TMUX is set and tmux is on PATH before sourcing this.
+
+_rd_winmode="${REVDIFF_TMUX_WINDOW:-}"
+if [ -z "$_rd_winmode" ]; then
+    # agent-deck markers: its env var (also mirrored into the tmux session env), with the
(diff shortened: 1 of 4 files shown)


paskal:
Thanks for the review — addressed in 1de0d10:

1. **EXIT trap from a sourced file.** Good catch. Removed the `trap … EXIT` entirely so it can't clobber the caller's cleanup trap. The backend now cleans the sentinel explicitly on every exit path, and in the non-agent-deck case returns *before* any trap/sentinel work, leaving the caller's environment untouched.

2. **Unbounded busy-wait.** `tmux new-window` failure is now handled (prints an error and exits non-zero instead of waiting on a sentinel that will never appear). The wait loop is bounded on the review window still existing — if the window disappears without writing the sentinel, we stop. A timer would be wrong here: a genuine review can legitimately take a long time, so window-existence is the correct bound.

3. **`sh -c` vs bash.** Intentional, for consistency: the tmux *popup* path runs `display-popup … -- sh -c "$REVDIFF_CMD"`, and every other backend (zellij/kitty/cmux/ghostty) runs the command through `sh`/`#!/bin/sh`. `REVDIFF_CMD` is built sh-compatible. Switching only this path to bash would be the divergent choice.

4. **Duplication across the two trees.** Follows the existing layout — `launch-revdiff.sh` itself is already maintained as two separate copies (`.claude-plugin/…` and `plugins/codex/…`), so the backend matches that convention. Happy to factor both into one shared source as a follow-up if you'd prefer, but that's a repo-wide change beyond this PR.

CI is green now — the failure was plain `shellcheck` (no `-x`) not following the runtime-resolved source (SC1091); silenced with a `disable=SC1091` directive on that line.

umputun (approved):
checked the update (1de0d10), all four addressed:

- EXIT trap gone, sentinel cleaned explicitly per path, the caller's `OUTPUT_FILE` trap stays in force.
- new-window failure exits 1 instead of hanging, and the wait loop also breaks if the window vanishes without a sentinel - covers more than the popup backends do.
- `sh -c` kept, consistent with the other backends and `REVDIFF_CMD` is sh-compatible.
- SC1091 silenced on the source line, CI green.

one cosmetic nit, non-blocking: the codex `agentdeck-window.sh` doesn't carry the `# source:` header the other codex script copies have. The two copies are byte-identical otherwise.

lgtm.

PR #233, opened 2026-06-20
feat(ui): Open focused source in editor

abhinav:
Add an `open_file_in_editor` action for the diff pane so reviewers can jump from the focused row to the current worktree file in their configured editor.

The default binding is `e`. `Ctrl+E` keeps the existing annotation-editor behavior.

Line navigation is best effort for worktree-backed reviews. Added and context rows request their current line, removed rows use the nearest current source line, and staged or ref reviews still request a line without reloading the displayed diff after a clean editor exit.

Demos:

<details>
  <summary>Opening an added line</summary>
  open-focused-file
</details>

<details>
  <summary>Opening a deleted line opens the nearest available line</summary>
  open-removed-line
</details>

<details>
  <summary>Opening a collapsed deleted line does not work</summary>
  open-collapsed-delete-placeholder
</details>


abhinav:
My bad on the lint failure. On it

umputun (changes requested):
couple issues to fix before merge:

1. `open_file_in_editor` should refuse to launch when the current file has line-level annotations. After the editor exits and the file reloads, annotations keyed by `(file,line,type)` can become orphaned or attach to the wrong line if the user deletes or moves annotated content. Since there is no stable anchor/rebase mechanism, blocking source editing for line-annotated files is safer than silently corrupting review comments. File-level annotations can still be allowed.

2. `sourceEditorFinishedMsg` should still capture the opened file. Completion currently calls `m.reloadCurrentFile()`, so it reloads whatever `m.file.name` is when the editor exits. If a queued file-load message changes the displayed file while the TUI is suspended, the file just edited may not be reloaded and another file may refresh instead. Capture the opened file and reload only that target, or reload only if it is still displayed.

3. `workDir != ""` is not a safe capability gate for this feature. `--stdin` is blocked, but compare mode, `--all-files`, standalone `--only`, and staged/ref reviews all have different source and reload semantics. Pls wire an explicit source-editor policy from the composition root instead of deriving support from `workDir` inside the UI. At minimum, document and test the supported modes: working tree reloads, staged/ref is best effort with no reload, compare opens `--compare-new`, stdin is unavailable.

4. pls reject source paths that escape the worktree, both lexically and after symlink resolution. `sourceEditorTarget` joins `workDir + file`, and `SourceCommand` uses `os.Stat`, which follows symlinks. A renderer path with `..` or a symlinked repo file can point outside the worktree and `e` would open that target in the editor.

minor:
- I'd avoid exposing `editor.SourceTarget` through the UI-side `ExternalEditor` interface. `SourceCommand(path string, line int)` or a UI-local request keeps the consumer-side interface cleaner.
- `app/ui/mouse.go:153` says mouse hint clearing mirrors `handleKey`, but mouse only clears reload/compact/editor hints while `handleKey` also clears key-chord and vim-motion hints. Either narrow the comment or match the behavior.

abhinav:
> open_file_in_editor should refuse to launch when the current file has line-level annotations.

Argh. I see the concern and I agree.
But that also has a slightly degraded experience: you're reviewing the file and then see a minor thing that's not worth round-tripping with the agent for, but can't edit.
But I agree that's still better than data loss.

If you're amenable to it, I'll start with a version where it's forbidden for annotated files but only if we're reviewing dirty changes. For staged, or for older commits or stdin, the diff output is stable and won't change. WDYT?

Will work on the others comments as well.

abhinav:
All comments addressed. Kept everything as separate commits for (hopefully) ease of review.

umputun (commented):
one remaining thing: pls sync the source-editor docs in the agent/plugin reference files with the updated README/site wording.

These still describe the old behavior and miss the current mode/annotation constraints:

- `.claude-plugin/skills/revdiff/references/usage.md`
- `plugins/codex/skills/revdiff/references/usage.md`

They should mention the same points as `README.md`: stable source path only, compare mode opens `--compare-new`, stdin is unavailable, and working-tree files with line annotations cannot be opened for editing.

abhinav:
@umputun updated as suggested.

Semi-related FYI: If you'd like to always keep some sections of these files in sync, I have a [small tool](https://github.com/abhinav/stitchmd/) I wrote some time ago to stitch larger .md files together from smaller ones. The idea is that sections get their own files, and then a separate file specifies the order to combine them in (similarly to [mdbook's SUMMARY.md files](https://rust-lang.github.io/mdBook/guide/creating.html#summarymd)). It just might fit into your setup.

umputun (commented):
round 2 looks good. All the round-1 items are addressed (line-annotation block, captured reload target, composition-root policy, path-escape guard, primitive interface, mouse comment), and the doc sync is done.

two minor things, neither blocking:

1. `docs/ARCHITECTURE.md` still references a type that no longer exists. Lines ~212/238/343 show `SourceCommand(target editor.SourceTarget)` / `SourceCommand(target)`, but the actual signature is `SourceCommand(path string, line int)` and `editor.SourceTarget` was removed in the interface flatten. Pls sync the doc.

2. optional hardening, not a blocker. The worktree-escape guard in `sourceEditorTarget` only runs for relative paths (inside `if !filepath.IsAbs`), so an absolute target skips both `filepath.IsLocal` and the `os.OpenRoot`/`root.Stat` check. Right now that's only reachable through user-named `--only` paths (VCS listings stay relative and do get validated), so it's not an escape of the untrusted surface. Just flagging that absolute targets aren't confined, in case you want symmetric validation later.

abhinav:
For absolute path escape—I think yeah right now that's working off of assumptions of current use.
Maybe, based on flags, we decide at entry point whether paths are allowed to escape root (if root is set)?
e.g. `--only=/tmp=foo` is good, but not `revdiff commita commitb` getting a malicious `diff` output.

abhinav:
Update: I'm going to hold off on folding that into this change unless you prefer that be handled here as well.

umputun (changes requested):
the ARCHITECTURE.md signature and the collapsed-delete rejection are both addressed correctly.

on the abs-path escape, agreed - let's hold it for this PR. The relative/untrusted-listing path is fully guarded, and an absolute target is only reachable through your own `--only` path, so symmetric validation can come later if at all.

one new thing, in the same area as the round-1 reload-target fix. `handleSourceEditorFinished` skips reload on `msg.fileName != m.file.name`, but `m.file.name` lags `m.tree.SelectedFile()` during a queued cross-file load. With `--cross-file-hunks`, a boundary `]`/`[` calls `m.tree.StepFile()` synchronously and queues the next-file load while `m.file.name` is still the old file. Press `e` in that window and the editor opens the old file; if that queued load is still in flight when the editor exits, `reloadCurrentFile()` bumps `loadSeq` and the next-file load is dropped as stale, so the displayed file and tree selection disagree. Narrow window and a non-default flag, no data loss, but it's the residual clobber the captured-target fix was meant to close. Anchoring the guard on `m.tree.SelectedFile()` (and clearing `pendingHunkJump` on mismatch) plus a regression test should cover it.

couple optional nits, take or leave:
- `sourceEditorTargetResult` is package-private but has exported fields with godoc-style comments - `editorFinishedMsg` right above it is unexported and lowercase, could match that.
- `SourceCommand`'s line-syntax switch has no default arm - a future mode constant added without an arm would build a command with no path. Latent, not a live bug.

also needs a rebase - it conflicts in `model.go` + `editor_test.go` now that #235 merged.

abhinav:
Fixed

> sourceEditorTargetResult is package-private but has exported fields

That was intentional but happy to unexport. Changed.

> SourceCommand's line-syntax switch has no default arm

added

umputun (approved):
round 3 items all addressed. Cross-file reload race now anchors the guard on the tree selection and clears the pending hunk jump, with a regression test covering the window. Unexport and default-arm nits done, rebase is clean.

lgtm, thx

PR #235, opened 2026-06-22
fix(ui): Restore mouse after editor return

abhinav:
Bug:
Ctrl+E annotation editing releases the terminal to the external editor.
Bubble Tea disables mouse tracking during that release,
then restores the terminal without re-emitting the mouse-enable sequence.
After returning from the editor,
events therefore stay with the terminal instead of reaching revdiff.

Fix:
Mark editor sessions that passed through Bubble Tea's process handoff,
and re-enable mouse cell tracking when those sessions complete
for sessions that started with mouse tracking enabled.

Editor setup failures retry without requesting mouse restoration,
and `--no-mouse` sessions do not gain mouse tracking after editor return.

### DIFF
--- a/app/main.go
+++ b/app/main.go
@@ -195,6 +195,7 @@ func run(opts options) (int, error) {
 		ReloadApplicable:  reloadApplicable(opts),
 		CompactApplicable: compactApplicable(opts, renderer),
 		NoColors:          opts.NoColors,
+		MouseTracking:     !opts.NoMouse,
 		NoStatusBar:       opts.NoStatusBar,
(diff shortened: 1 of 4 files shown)


abhinav:
This change is part of the following stack:

- #235 ◀

<sub>Change managed by [git-spice](https://abhinav.github.io/git-spice/).</sub>

umputun (approved):
lgtm. Verified the diagnosis against vendored bubbletea - `RestoreTerminal` re-enables altscreen/bracketed-paste/report-focus but never mouse, so the fix targets the real gap, and gating on `mouseTracking` keeps `--no-mouse` correct.

PR #236, opened 2026-06-23
feat: add --watch flag for automatic reload on file changes

Lysander66:
Polls ChangedFiles() every 2s; reloads silently when the file list hash changes. Skips reload when annotations exist to avoid data loss. Disabled in --stdin and --compare modes.


umputun:
going to decline and close this, with the reasoning laid out rather than a silent close.

the core issue is there's no problem being solved here. The description explains the mechanism (poll every 2s, reload when the file-list hash changes) but never says what it is for or who needs it. No linked issue, no use case, so there is nothing to weigh the design against.

even taking the most likely intent, "refresh the diff without pressing a key", revdiff already reloads on `R`. So `--watch` doesn't add a capability, it just auto-fires the existing reload on a timer.

and it doesn't do what the title says. `fileListHash` hashes only path and status per entry, so editing a file that is already in the diff (the common case) changes neither and watch stays silent. It only reloads when a file enters or leaves the changed set or flips status. A flag called "automatic reload on file changes" that ignores edits to changed files isn't reloading on the changes people would expect.

other problems on top of that:
- no tests. 80 lines including a polling goroutine, a new message type and a new handler, none of it covered.
- no docs. A new flag here has to be reflected in README, site/docs.html and the plugin reference files.
- the watcher goroutine calls the renderer's `ChangedFiles` while the model calls the same renderer from its own goroutines, which is a data-race risk that isn't addressed.

if there is a real workflow that the `R` reload doesn't cover, open an issue describing it first and we can talk through whether auto-watch fits and how it should behave before any code.

Lysander66:
[diffnav](https://github.com/dlvhdr/diffnav) ships --watch for exactly this use case.

  The fix is straightforward: hash `git diff HEAD --stat` output instead, which
  changes on content edits. That also eliminates the renderer concurrency concern
  since the watcher runs its own subprocess.

  Happy to leave it here since you've closed the PR.

PR #237, opened 2026-06-23
Add automatic light/dark theme selection

jrpat:
## Problem

Many users alternate between light and dark environments and use dark backgrounds sometimes and light backgrounds sometimes. Frequently the OS is set up to automatically switch between the two.

Revdiff currently only has a way to configure a single theme which means revdiff startup can often be jarring, especially if system is dark and revdiff is light.

I am one of these users and frequently do a "gah my eyes!" the first time I run revdiff after my system switches to dark mode.

## Solution

This adds a new special theme called `auto` and 2 new config flags `auto-theme-light` and `auto-theme-dark`. If `theme = auto`, we automatically resolve to either the light or dark theme depending on the terminal background color.

This is analogous to a similar feature in [hunk](https://github.com/modem-dev/hunk#config).

## ps

💛 revdiff is fantastic. Thanks for it!


jrpat:
Sorry I missed the "if this is your first contribution, start with an issue" instruction. Feel free to close this if you're not interested. I implemented it to scratch my own itch, so no wasted time 😉

PR #247, opened 2026-06-30
feat: add a Nix flake

### LINKED ISSUE #230: add flake.nix
maybe even publish to nixpkgs

mvanhorn:
## Summary

Adds a minimal `flake.nix` (and its `flake.lock`) so revdiff can be built, run, and hacked on through Nix:

- `nix run github:umputun/revdiff -- <args>` runs the TUI with no manual Go toolchain setup
- `nix build` produces the `revdiff` binary at `result/bin/revdiff`
- `nix develop` drops into a shell with Go available

## Why this matters

Requested in #230, where you confirmed you were "happy to take a minimal flake.nix as a PR" and sketched the exact recipe this change follows. Packaging revdiff as a flake lets Nix users install and pin it alongside the rest of their tooling, and `nix run github:umputun/revdiff` gives a zero-install way to try it.

The flake stays in lockstep with the existing build:

- `buildGoModule` with `vendorHash = null` builds straight from the committed `vendor/` tree, so the build needs no network access and tracks `go mod vendor`.
- `CGO_ENABLED = 0` and `ldflags = [ "-s" "-w" "-X main.revision=${version}" ]` mirror the Makefile, so `revdiff --version` is populated.
- The binary built from `./app` is renamed to `revdiff`, and `meta.mainProgram = "revdiff"` makes `nix run` resolve it.

Per your request the change is limited to `flake.nix` + `flake.lock`, with no CI changes and nixpkgs packaging left to the community.

## Testing

- `go build ./...` passes against the vendored tree.
- `flake.lock` pins `nixpkgs` (nixpkgs-unstable), `flake-utils`, and flake-utils's `systems` input to released revisions; the pinned nixpkgs defaults to Go 1.26, matching `go 1.26` in `go.mod`.
- `doCheck = false` skips the Go test phase during the build, since some tests (e.g. VCS detection) expect a `.git` working tree that Nix's source copy omits.

Fixes #230

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -1,0 +1,61 @@
+{
+  "nodes": {
+    "flake-utils": {
+      "inputs": {
+        "systems": "systems"
+      },
+      "locked": {
+        "lastModified": 1726560853,
+        "narHash": "sha256-X6rJYSESBVr3hBoH0WbKE5KvhPU5bloyZ2L4K60/fPQ=",
+        "owner": "numtide",
+        "repo": "flake-utils",
+        "rev": "c1dfcf08411b08f6b8615f7d8971a2bfa81d5e8a",
+        "type": "github"
+      },
+      "original": {
+        "owner": "numtide",
+        "repo": "flake-utils",
+        "type": "github"
+      }
+    },
+    "nixpkgs": {
+      "locked": {
+        "lastModified": 1781607440,
+        "narHash": "sha256-rxO+uc/KFbSJp+pgyXRuAX6QlG9hJdnt0BXpEQRXY+U=",
+        "owner": "NixOS",
+        "repo": "nixpkgs",
+        "rev": "3e41b24abd260e8f71dbe2f5737d24122f972158",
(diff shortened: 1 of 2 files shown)


umputun (approved):
lgtm. Matches the #230 recipe - ldflag hits the real `main.revision` symbol, license/vendorHash/eachDefaultSystem all check out, and the three lock inputs are official upstreams.

two notes, neither blocking:

1. couldn't run `nix build` here to confirm it actually builds. Worth a quick `nix build` + `nix run . -- --version` before merge - `go.mod` is `go 1.26` with no `toolchain` directive, so the pinned nixpkgs `pkgs.go` has to be >= 1.26 or the sandboxed build hard-fails with no toolchain fallback.

2. `version = self.shortRev or ...` makes `revdiff --version` print a bare commit hash on nix builds (`dev` from a non-git tarball) instead of the `tag-hash-date` from goreleaser/make. Fine to leave, pure flakes can't read git tags cleanly, just flagging.

PR #256, opened 2026-07-11
Reset reviewed files when their diffs change

melonamin:
## Summary

- store a semantic diff fingerprint with each reviewed file
- revalidate only reviewed files during file-list reloads
- preserve reviewed state across line-number and surrounding-context shifts
- discard stale asynchronous mark results across reload generations

## Why

Reviewed state was keyed only by path. Reloading after a new commit or rebase could leave a changed file marked reviewed.

Fingerprints use file metadata and ordered added/removed content while ignoring line numbers, context rows, and compact-mode dividers. Changed, removed, or unreadable files lose their reviewed mark; unchanged files retain it.

Implements the behavior proposed in https://github.com/umputun/revdiff/discussions/251.


umputun (commented):
no blockers here, the change is sound. Tests/lint/race green on my side and the async sequencing holds up under tracing. A few things worth cleaning up, none critical.

**dead code left by the refactor.** `fetchEffectiveFileDiff` now owns the staged-retry and untracked-read fallback, and `loadFileDiff` sets `resolved: true`, so `resolveEmptyDiff`, the `fileLoadedMsg.resolved` field, and the `if !msg.resolved` guard in `handleFileLoaded` never run in production. Only the older tests still build a `fileLoadedMsg` without `resolved`. No runtime effect, but the same fallback now lives in two places. I'd drop `resolveEmptyDiff` plus the field and guard; the raw-`fileLoadedMsg` tests all pass non-empty lines, so they keep passing.

**binary files can keep a stale ✓.** A binary diff has no add/remove rows, so `FileFingerprint` hashes the placeholder text. On hg/jj that's a bare `(binary file)` for any change, so a reviewed binary stays marked after reload regardless of what actually changed; on git it carries a size, so only a same-size swap collides. It's advisory (nothing more to inspect for a binary in the TUI) so minor, but it's the one spot where the mark can be wrong. The `IsBinary`/`IsPlaceholder` branch is also currently untested.

**async mark can be lost on a file that survives the reload.** Mark a not-yet-loaded tree file while an `R` reload is in flight: if the `filesLoadedMsg` is handled before the fingerprint result, `handleFilesLoaded` clears the whole `pending` map (loaders.go:386) and the mark is dropped; if the fingerprint arrives first it sticks. The wipe is there to stop a removed path coming back, which is right, but making it path-aware (drop only pending entries missing from the new `entries`) keeps the surviving ones. Narrow window and re-pressable, so minor. Related: a pending mark shows no ✓ until it resolves and a second space silently cancels it, so a bit of feedback there wouldn't hurt.

**smaller stuff.**
- `else if` after a `return` in the staged-retry (loaders.go:247); the rest of that func is flat early-returns.
- test coverage: the 4-worker path never runs with more than one reviewed file, the per-path stale-seq guard and the error branch in `handleReviewFingerprintLoaded` are uncovered, and the strict untracked-read error path is untested. `TestModel_LoadFilesFingerprintsOnlyReviewedPaths` also appends to a shared slice with no lock, which would race under `-race` the moment a second reviewed path is added.
- docs: the ✓ now clears on reload when a file's diff changed, which is user-visible, but README/docs.html/usage.md and the CLAUDE.md gotchas + ARCHITECTURE notes aren't updated.

**on the premise.** I never replied on #251, so for the record: the diff-fingerprint approach is reasonable and matches GitHub's "viewed", I'm good with the direction. One property to be aware of: since the fingerprint drops line numbers and context, an identical edit moved to another spot with the same surrounding lines keeps its mark. That's the intended tradeoff for surviving rebases, just noting it.

umputun (commented):
round 2: all of it addressed and clean here, build/test/lint/race green. The path-aware `pending` wipe, the `ReviewFingerprintStable` binary guard, dropping `resolveEmptyDiff` plus the `resolved` field, and the added tests all look right.

one minor cleanup for later: `plugins/codex/skills/revdiff/references/usage.md` still has the old `✓ Space` row and is missing the new reviewed-mark paragraph. The `.claude-plugin` copy got both, so the codex mirror needs the same edit to stay in sync. No rush, fine to do after.

PR #257, opened 2026-07-12
feat(ui): filter unreviewed files

melonamin:
## Summary

- add an `F` toggle for showing only unreviewed files
- advance to the next file as the current one is marked reviewed
- keep annotated and unreviewed filters separate

This builds on the smarter reviewed-state reconciliation from #256. Reviewed marks now stay trustworthy across reloads, so filtering by them is useful during large reviews.

## Checks

- `TMPDIR=/dev/shm go test ./...`
- `golangci-lint run`
- Codex review: no actionable findings

### DIFF
--- a/.claude-plugin/skills/revdiff/references/usage.md
+++ b/.claude-plugin/skills/revdiff/references/usage.md
@@ -31,7 +31,7 @@ some-command | revdiff --stdin --output /tmp/annotations.txt      # annotate gen
 
 ## Single-File Mode
 
-When a diff contains exactly one file, revdiff automatically hides the file tree pane and gives full terminal width to the diff view. Pane-switching keys (`Tab`, `h/l`, `n/p`, `f`) become no-ops, except when markdown TOC is active (see below). Search navigation (`n`/`N`) still works normally.
+When a diff contains exactly one file, revdiff automatically hides the file tree pane and gives full terminal width to the diff view. Pane-switching keys (`Tab`, `h/l`, `n/p`, `f`, `F`) become no-ops, except when markdown TOC is active (see below). Search navigation (`n`/`N`) still works normally.
 
 ## Markdown TOC Navigation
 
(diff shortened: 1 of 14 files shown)


umputun (changes requested):
mechanically the wiring is clean, tests/lint/race green and the filter mutual-exclusivity, empty-state, single-file, and async paths all check out. But the auto-advance itself has a real bug that fires on any repo with subdirectories, so i'd fix at least the first one before merge. Both are non-destructive (wrong file selected, nothing lost) and both come from `selectAfterRebuild` walking source order.

**auto-advance selects the wrong file (deterministic, any multi-dir repo).** `selectAfterRebuild` scans `ft.allFiles[previousIdx+1:]`, which is git/renderer order, but the tree displays `buildEntries` order (dir-grouped, sorted). So "next unreviewed" is next-in-source-order, not next-on-screen. e.g. source `[a/c.go, b.go, a/d.go]` displays `[b.go, a/c.go, a/d.go]`; reviewing b.go selects a/d.go and skips a/c.go, which sits right where b.go was. This repo's own tree (root files + app/… subdirs) hits it. The new tests only use flat same-dir files (`a.go`/`b.go`/`c.go`) where source == display order, so they don't catch it. Fix: walk `ft.entries` (display order) forward from the reviewed file's display position instead of `allFiles`.

**reload arriving mid-advance jumps the selection backward (race).** With the filter active, if a reload's `filesLoadedMsg` is handled after you mark a file but before its auto-advance load arrives, `handleFilesLoaded` does `Rebuild` (resets the cursor to the top), then the `ReconcileReviewed` refresh settles on the first unreviewed file, then `SelectByPath(m.file.name)` can't restore anything because `m.file.name` still points at the just-reviewed, now-hidden file. So the selection ends on the first unreviewed file instead of the intended next one, and the follow-up load bumps `file.loadSeq` and drops the pending advance. Narrow (needs an unreviewed file before the marked one, plus the reload winning the goroutine race) and non-destructive. The display-order fix above reduces the surprise; fully fixing it means capturing the pre-`Rebuild` selection so the reconcile advances relative to it, not the reset-to-top cursor.

no test covers the reload-completes-between-mark-and-advance-load ordering; worth adding one if you touch this.

umputun (approved):
both fixed, verified here. The advance now walks `ft.entries` (display order) via `nextUnreviewedAfterCursor`, and the reload path captures the pre-`Rebuild` selection so the auto-advanced file survives. The two new tests cover exactly the cases that were broken: multi-dir advance order, and a reload arriving mid-advance. Build/test/lint/race green. Good to merge.

PR #259, opened 2026-07-13
fix(ui): preserve unreviewed tree scroll

### LINKED ISSUE #258: Marking a file reviewed while filtering unreviewed files jumps the tree scroll position
### What is the problem?

After the unreviewed-files filter added in #257 is enabled with `F`, pressing `Space` to mark the selected file as reviewed removes it from the filtered tree and advances the selection, but also scrolls the tree so the next selected file is on the last visible row.

This is especially disruptive after skipping many files and starting review from the middle or bottom of the tree. Even when the reviewed file was at the top or middle of the viewport, each review action jumps the next file to the bottom, making it harder to retain visual context.

### Steps to reproduce

1. Open a diff containing enough files for the file tree to scroll.
2. Press `F` to show only unreviewed files.
3. Skip several files so the selection is somewhere in the middle or lower part of the tree.
4. Press `Space` to mark the selected file as reviewed.

### Actual behavior

The reviewed file disappears and the tree scrolls, placing the next selected file on the last visible row.

### How does this solve it?

Preserve the selected row’s approximate viewport position when the reviewed file is removed from the filtered tree. The next selected file should stay where the reviewed file was, unless scrolling is actually required to keep it visible. This removes the unnecessary jump while retaining the automatic advancement introduced in #257.

### Version

v1.11.0

mishamsk:
- Keep the next unreviewed file on the same visible row after marking the current file reviewed.
- Preserve the unreviewed tree scroll position across force reloads, resetting it when the filtered tree becomes empty.

Closes #258.

### DIFF
--- a/app/ui/loaders.go
+++ b/app/ui/loaders.go
@@ -399,20 +399,7 @@ func (m Model) handleFilesLoaded(msg filesLoadedMsg) (tea.Model, tea.Cmd) {
 			delete(m.reviewed.pending, path)
 		}
 	}
-	// Preserve the visible selection across a reload. In unreviewed-only mode
-	// the displayed diff may still be the file that was just hidden while its
-	// auto-advance load is in flight, so m.file.name cannot restore the intended
-	// next selection below.
-	selectedBeforeRebuild := ""
-	selectedVisibleRow := 0
-	if m.tree.UnreviewedFilterActive() {
-		selectedBeforeRebuild = m.tree.SelectedFile()
-		selectedVisibleRow = m.tree.SelectedVisibleRow()
-	}
 	m.tree.Rebuild(entries)
-	if selectedBeforeRebuild != "" {
-		m.tree.SelectByPathAtVisibleRow(selectedBeforeRebuild, selectedVisibleRow)
-	}
 	m.tree.ReconcileReviewed(msg.reviewedBefore, msg.reviewedFingerprints)
 	m.reviewed.cache = make(map[string]string, len(msg.reviewedFingerprints))
 	maps.Copy(m.reviewed.cache, msg.reviewedFingerprints)
--- a/app/ui/loaders.go
+++ b/app/ui/loaders.go
@@ -404,12 +404,14 @@ func (m Model) handleFilesLoaded(msg filesLoadedMsg) (tea.Model, tea.Cmd) {
 	// auto-advance load is in flight, so m.file.name cannot restore the intended
 	// next selection below.
 	selectedBeforeRebuild := ""
+	selectedVisibleRow := 0
 	if m.tree.UnreviewedFilterActive() {
 		selectedBeforeRebuild = m.tree.SelectedFile()
+		selectedVisibleRow = m.tree.SelectedVisibleRow()
 	}
 	m.tree.Rebuild(entries)
 	if selectedBeforeRebuild != "" {
-		m.tree.SelectByPath(selectedBeforeRebuild)
+		m.tree.SelectByPathAtVisibleRow(selectedBeforeRebuild, selectedVisibleRow)
 	}
 	m.tree.ReconcileReviewed(msg.reviewedBefore, msg.reviewedFingerprints)
 	m.reviewed.cache = make(map[string]string, len(msg.reviewedFingerprints))
--- a/app/ui/loaders_test.go
+++ b/app/ui/loaders_test.go
@@ -1867,22 +1867,32 @@ func TestModel_FilesReloadPreservesUnreviewedScrollOffset(t *testing.T) {
 	}
 	m := testModel(nil, nil)
 	m.tree.Rebuild(entries)
+	m.tree.SetReviewed("a.go", "fp-a")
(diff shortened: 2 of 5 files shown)


umputun (commented):
core fix is right and does what #258 asked. Verified the reload test fails without it (offset clamps to 2 instead of 4), and build/test/lint/race are all green locally. Two viewport nits, one of them a small regression this PR adds.

**`F` toggle scrolls the `./` header off-screen (should fix before merge).** `selectAfterRebuild` (app/ui/sidepane/filetree.go:445) no longer resets `offset` in its no-anchor fallback, and `ToggleUnreviewedFilter` reaches that fallback with both args empty. So toggling `F` while the tree is scrolled down leaves `offset` clamped to the first file's index (1) instead of 0, and the leading `./` directory header gets scrolled off the top. It's the milder version of the same jump #258 was about. The mark-reviewed and reload paths return early via `SelectByPath`, so resetting in the fallback doesn't touch the fix:

```go
	if preferred != "" && ft.SelectByPath(preferred) {
		return
	}
	ft.offset = 0 // no anchor survived, reset viewport to top
	for i, e := range ft.entries {
		if !e.isDir {
			ft.cursor = i
			return
		}
	}
```

worth a small test for the toggle-from-scrolled case, none of the new ones exercise it.

**offset is preserved by index, not visible row (optional).** on a reload that changes the unreviewed set *above* the current viewport, the selected file's row drifts by the number of rows added or removed above it, since `offset` is kept by absolute index rather than by visible row (`cursor - offset`). It's rare, it's outside the mark-reviewed case (#258 is exact there), and it's still better than the old full reset, so fine to defer. If you want it airtight later: capture `visibleRow = cursor - offset` before `Rebuild` in `handleFilesLoaded`, then set `offset = clamp(newCursor - visibleRow)` once the selection is restored.

mishamsk:
Addressed both viewport points. The resulting behavior is:

- Marking the selected file reviewed in `F` view removes it and keeps the next unreviewed file on the same visible row.
- Reloading with `R` in `F` view keeps the surviving selected file on the same visible row.
- Consecutive `R` reloads no longer drift the selection toward the top.
- If a reload adds or removes filtered files above the selection, the selected file is restored by path and remains on the same visible row rather than preserving a stale absolute offset.
- Switching from the unfiltered view into `F` view makes a best effort to preserve both the selected file and its visible row when that file remains unreviewed.
- If no selection anchor survives when entering `F` view, the tree falls back to the first visible file with offset 0, keeping the `./` header visible.
- An empty unreviewed view resets its offset to 0.
- Switching from `F` view back to the unfiltered view still resets the viewport to the top, matching current `master` behavior. This is a little asymmetric with the new best-effort preservation in the other direction; happy to make the reverse toggle preserve position in a follow-up if that is preferable.
- Reload behavior outside `F` view is unchanged.

Manually smoke-tested both entering `F` from a scrolled middle selection and multiple consecutive `R` reloads with reviewed files filtered above the selection. Automated race/coverage tests and lint are also green.

umputun (commented):
both viewport points from last round are handled and the fix itself is correct. Build/test/lint/race are green, and the visible-row anchoring (`offset = max(cursor - visibleRow, 0)`) is the exact inverse of the capture, so the row survives index shifts. One thing still open.

**the `./`-header regression fix has no test.** The reset is in place: `selectAfterRebuild` zeroes `offset` on the non-empty no-anchor fallback at `app/ui/sidepane/filetree.go:459`. But nothing reaches that path from a scrolled state. `TestFileTree_EnableUnreviewedFilterPreservesVisibleSelection` keeps the selected file unreviewed, so toggling `F` takes the anchored branch at line 452, not 459. `TestFileTree_UnreviewedFilterEmptyStateResetsOffset` hits the separate `len(entries)==0` early return at 447. So deleting line 459 leaves the whole suite green, and `EnsureVisible` won't cover for it: after the reset the cursor sits on the first file at index 1, below the `./` header at index 0, so a stale offset still scrolls the header off. This is the toggle-from-scrolled case I flagged last round. Add a `filetree_test.go` case that reaches 459 from `offset > 0` (e.g. select a reviewed file, then `ToggleUnreviewedFilter` with entries remaining) and asserts `offset == 0` and the first entry visible.

optional, take or leave:

- `selectAfterRebuild` repeats `ft.offset = max(ft.cursor-visibleRow, 0)` verbatim in the `previous` and `preferred` branches (452, 456). One `if a || b { ... }` covers both.
- `visibleRow := ft.cursor - ft.offset` is re-derived at three call sites (`Rebuild`, `ToggleUnreviewedFilter`, `RefreshUnreviewedFilter`). A small `ft.visibleRow()` method would name the concept and stop it drifting.
- when `ft.unreviewed`, `Rebuild` anchors against the full unfiltered list even though `RefreshUnreviewedFilter` rebuilds the filtered view right after. It's load-bearing only because it carries the `cursor - offset` delta forward, and nothing says so. A one-line comment there would keep a future `Rebuild` cleanup from silently dropping scroll preservation.

PR #261, opened 2026-07-15
fix(themes): detect terminal background inside tmux for --theme auto

kylesnowschwartz:
**Disclosure: this is an AI-generated proposal** (written by Claude in a coding session I was driving). I reviewed the code and QA'd the behavior by hand, and it fixed the problem on my machine, but please review it as machine-written code.

`--theme auto` always picked the dark theme inside tmux. termenv refuses to send OSC status-report queries when `TERM` starts with `tmux` or `screen` (vendor/github.com/muesli/termenv/termenv_unix.go, the multiplexer guard in `termStatusReport`), so the background query never ran. Detection fell through to `COLORFGBG` (usually unset) and then the dark default, no matter what the terminal looked like.

Detection now runs two tmux-aware paths before giving up:

1. `tmux display-message -p '#{client_theme}'` (tmux >= 3.5). tmux reports the outer terminal's light/dark theme from the terminal's native theme reporting. This needs no tty, so it also works inside `display-popup` panes, where tmux does not answer OSC 11 at all. The subprocess has a 500ms timeout so an unresponsive tmux server falls through to the next path instead of stalling startup.
2. termenv's own OSC 11 query, with `TERM` masked as `xterm-256color` via a custom `termenv.Environ` so the multiplexer guard does not trip. tmux answers OSC 11 in regular panes with the outer terminal's current background. On older tmux this covers panes; popups fall through to the existing default.

Outside tmux, behavior is unchanged: plain `termenv.HasDarkBackground()`. If both tmux paths fail, the old fallback still applies (`COLORFGBG`, then dark).

QA'd on macOS with tmux 3.7b and Ghostty, light terminal theme active:

- regular pane: `--theme auto` resolves the light theme (before: always dark)
- `display-popup`: resolves the light theme (before: always dark; this is the path the Claude/codex plugin launchers use)
- switching the terminal between light and dark mid-session changes the result without reattaching, since Ghostty pushes theme-change reports and tmux tracks them
- outside tmux: unchanged

One known limit: with several clients attached to one session from terminals with different themes, tmux resolves `#{client_theme}` against its notion of the current client, which can be the other terminal. The OSC 11 fallback is relayed through the same client choice, so neither path is worse than the other here.

### DIFF
--- a/app/termbg.go
+++ b/app/termbg.go
@@ -1,0 +1,116 @@
+package main
+
+import (
+	"context"
+	"os"
+	"os/exec"
+	"strings"
+	"time"
+
+	"github.com/muesli/termenv"
+)
+
+// maskedTerm is the TERM value reported to termenv when the real TERM belongs
+// to tmux. any plain xterm value works — it only needs to not match termenv's
+// multiplexer prefixes so the OSC query path stays enabled.
+const maskedTerm = "xterm-256color"
+
+// tmuxQueryTimeout bounds the tmux subprocess so a wedged tmux server (e.g.
+// blocked on a run-shell hook) degrades to the next detection method instead
+// of stalling startup indefinitely. display-message answers over the local
+// socket in single-digit milliseconds when healthy.
+const tmuxQueryTimeout = 500 * time.Millisecond
+
+// detectDarkBackground reports whether the terminal background is dark-ish.
+// termenv refuses to send OSC status-report queries when TERM starts with
+// "tmux" or "screen" (multiplexers historically swallowed them), so inside
+// tmux its detection never queries the terminal and always falls back to
+// "dark". two tmux-specific paths run first:
+//
+//  1. `tmux display-message -p '#{client_theme}'` (tmux >= 3.5) — fed by the
+//     outer terminal's native light/dark reporting, needs no tty, and works
+//     in display-popup panes where tmux does not answer OSC 11 at all.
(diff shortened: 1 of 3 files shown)


umputun (approved):
lgtm. Traced the premise against vendored termenv: masking `TERM` to `xterm-256color` does re-enable the OSC path, since `termStatusReport` only bails on `screen`/`tmux`/`dumb` prefixes (termenv_unix.go:238). The non-tmux branch is byte-identical to the old `HasDarkBackground()` call, so no regression there. Build, tests, lint, and race all clean.

one minor thing, not blocking:

the PR body says the 500ms timeout means an unresponsive tmux server falls through instead of stalling startup. That only covers path 1, the `tmux display-message` subprocess. Path 2 at `app/termbg.go:45` (the masked OSC 11 query) is bounded only by termenv's internal 5s `OSCTimeout`. So on old tmux (< 3.5, no `#{client_theme}`) inside a display-popup, if the surface answers neither OSC 11 nor the `CSI 6n` cursor query, `--theme auto` startup can stall up to ~5s before falling back to COLORFGBG/dark.

in practice the `CSI 6n` cursor-position answer short-circuits it (termenv_unix.go:281-283) and tmux does answer that, so realistic runs return fast. But the popup path is exactly the plugin-launcher case, so it's worth either noting the old-tmux latency ceiling or giving path 2 its own shorter deadline.

PR #263, opened 2026-07-15
feat(launch): run tmux reviews in a detached session so the popup can be backgrounded

kylesnowschwartz:
> **Draft** — see Testing for what still blocks it.

## Why

The tmux popup is all-or-nothing. Mid-review I often need the terminal — run a command, check a file, answer the agent — and closing the popup kills revdiff along with every annotation in it.

## What

revdiff runs in a detached tmux session and the popup is only a client attached to it. `prefix d` closes the popup and backgrounds the review; `launch-revdiff.sh --resume` re-opens the newest backgrounded review and returns annotations on the usual exit codes. Resume reads what it needs off the tmux session, so it also works after the original launcher is gone — the normal case once an AI harness hits its command-timeout cap. Both skills tell agents to reach for `--resume` when a user says they dismissed the popup, instead of relaunching and abandoning the annotations already made.

Because popup exit no longer means the review finished, completion is signalled by a sentinel file — the same pattern the zellij and kitty backends already use.

Two changes are independent of the backend question, and were called out in review as worth having either way:

- **Overlay stderr capture.** A fast-failing revdiff (bad ref, a flag an older installed binary does not know) used to produce a bare exit 1, because the overlay closed before anyone could read the error. Every sentinel backend now captures stderr and the launcher replays it on failure.
- **`REVDIFF_TMUX_SESSION_OPTIONS`.** Whitespace-separated `key=value` tokens applied to the review session before any client attaches, so external session tooling can recognize and skip it. A bad token warns and the review continues.

## Relationship to window mode (#266)

#266 landed `REVDIFF_TMUX_WINDOW=1` after this branch was written, and it also survives a client disconnect. Worth stating plainly: **disconnect resilience is no longer this PR's contribution.** What differs is where the review renders and how a reviewer steps away from it.

|  | window mode (master) | this PR |
| --- | --- | --- |
| tmux object | `new-window` in the current session | `new-session -d`, a separate hidden session |
| renders as | a full-size window beside your work | a floating popup over the current window |
| window list | gains a `review: ...` entry | untouched |
| sizing | terminal size | `REVDIFF_POPUP_WIDTH` / `REVDIFF_POPUP_HEIGHT` |
| stepping away | switch windows | `prefix d` — popup closes, you are back where you were |
| coming back | switch to the window | `launch-revdiff.sh --resume` |

The two coexist rather than compete: `REVDIFF_TMUX_WINDOW=1` bypasses everything here, `=0` forces the popup path.

Not changed: the plan-review launcher still runs revdiff inside the popup. It creates no session, so there is nothing to background.

## Testing

- `go test ./app` and `shellcheck -x` clean. The fake tmux backend models the session commands, and `TestDiffLauncherResumeRejectsUnusableInvocations` covers the `--resume` guards on both launcher copies.
- `--resume` in an isolated tmux sandbox: 12/12, stable across three runs. Notably, with the original launcher still blocked on the same sentinel, both it and the resume client printed the same annotations and exited 10 — a review is reported once per waiter, never consumed out from under the other.
- **Not re-run since the rebase onto post-#266 master: the live and signal QA** (full review in the popup, detach/reattach, ctrl-c cancel, SIGTERM mid-review). That is what keeps this in draft.

Split out of this PR: #291, which drops the skills' pre-flight tool checks and is unrelated to the backend.


umputun:
saw this is still a draft, so just a heads-up in case you haven't caught the merge: #266 went in yesterday and overlaps a good chunk of this. Flagging so you don't sink more effort into parts that are already in master.

what it covers:
- annotation-loss on disconnect: the binary now saves annotations to history on SIGHUP/SIGTERM (a `shutdownGuard` in `app/signal.go`), and drains SIGINT so a Ctrl-C meant for `$EDITOR` doesn't quit. Same ground as your SIGHUP / save-on-error commit.
- disconnect survival: `REVDIFF_TMUX_WINDOW=1` runs the review in a server-owned `tmux new-window` (see `agentdeck-window.sh`) that survives an SSH/VPN drop, reattach and the live review is still there.

what it does not cover, in case these are still worth it to you:
- your actual popup UX: keep the popup, detach it, review keeps running, reattach to the same popup. #266 uses a separate window, not a backgroundable popup, so that specific experience isn't in.
- `REVDIFF_TMUX_SESSION_OPTIONS` for tagging the transient session.
- capturing revdiff's stderr to `<sentinel>.err` and replaying it on non-0/10 exit. That diagnostic gap is real regardless of backend.

no rush and no pressure either way, just didn't want you polishing something that's already partly in master.

kylesnowschwartz:
Thanks for the heads-up, it saved me from polishing work that #266 already covers. I rebased this branch onto master and dropped the SIGHUP/save-on-error commit, since shutdownGuard handles both cases with tests. What's left here is the backgroundable popup, REVDIFF_TMUX_SESSION_OPTIONS, and the stderr capture.

I'm going to drive the new master with REVDIFF_TMUX_WINDOW=1 for a while before deciding what to do with this PR. If window mode covers my day-to-day cases I'll slim this down to the stderr capture and session options; if I miss the popup I'll mark it ready as is.

kylesnowschwartz:
Closing this. Window mode covers what I wanted from the backgroundable popup: I can move away from the review window and come back to it, so detaching a popup and reattaching to it is not something I need.

Of the three things left on this branch after the rebase, the stderr capture landed separately in #292, and the backgroundable popup is the part I am dropping. That leaves `REVDIFF_TMUX_SESSION_OPTIONS`, which existed so a session manager could recognize and skip the transient session this branch created. Window mode creates no session, so there is nothing to tag.

Thanks for the earlier heads-up about #266. It stopped me from finishing work that was already in master.

Issue #179, opened 2026-05-11
Cannot interrupt an active mouse scroll with an opposite scroll


Nihisil:
Version: v1.1.1

Steps to reproduce:
* Open a large file.
* Scroll down using the mouse wheel a few times.
* While the file is still scrolling downward, try to scroll up.

Current behaviour:
* The downward scroll event fully completes first.
* Only after that does the upward scroll event begin.

Expected behaviour:

* The downward scroll event should be interrupted.
* The upward scroll event should start immediately.

^this expected behaviour is based on standard conventions in other TUI/GUI applications

umputun:
fixed in #180, shipped in v1.2.0.

diff-pane wheel events are now coalesced so a reverse-direction wheel reacts fast instead of queueing behind a backlog of renders.

https://github.com/umputun/revdiff/releases/tag/v1.2.0

PR #265, opened 2026-07-15
feat(ui): off-screen change-count labels in diff frame (--frame-labels)

fkirill:
Off-screen change-count hints in the diff pane's frame borders, gated behind `--frame-labels` (off by default).

- Bottom border: `+x/-y modified lines below` — changes past the viewport bottom
- Top border: `+x/-y modified lines above` — changes scrolled above the viewport top
- Collapses to `no modified lines below` / `no modified lines above` at the ends

`+x` uses the add color, `-y` the remove color, the rest the frame color. Updates live on scroll. Skipped for pure-context views (plain file, markdown TOC) — no changes to count.

```
┌─ +4/-3 modified lines above ───────────────────────────────┐
│  ...diff...                                                │
└─ no modified lines below ──────────────────────────────────┘
```

## Motivation

In a large file you can't tell whether more changes lie below the viewport without scrolling to the end. `no modified lines below` says the rest of the file is unchanged — skip the scroll, move to the next file. The top label does the same looking back: `no modified lines above` means nothing was missed higher up.

## Flag

`--frame-labels` / `REVDIFF_FRAME_LABELS` / `frame-labels = true`. Off by default.

## Implementation

- `app/ui/framelabel.go` — post-render surgery on the pane's top/bottom border lines (sibling of the scrollbar transform in `scrollbar.go`). Reuses the line's own leading SGR to restore frame color after each colored run, so display width and the ANSI envelope stay intact (safe for the horizontal pane join).
- `modifiedAbove` / `modifiedBelow` map the viewport edge to a diff-line index via the existing `visualRowToDiffLine`, then `diff.CountChanges` on the slice above/below.
- Wired into both `View()` branches after `applyScrollbar`; gated by `frameLabelsEnabled()` (flag + file loaded + has changes).
- Flag threaded `options` → `main.go` → `ModelConfig.FrameLabels` → `modelConfigState.frameLabels`. Model reads the resolved bool, does not re-derive.

## Not included

- No SKILL.md entry: display-only preference, not agent-actionable — matches how `--vim-motion` / `--line-numbers` / `--word-diff` / `--compact` are handled (references only, absent from SKILL.md).
- No runtime toggle key — flag/env/config only.

## Docs

Flag row added to README.md, site/docs.html, references/config.md.

## Tests

`app/ui/framelabel_test.go` — above/below counts (top/mid/end scroll), top+bottom label build, no-changes states, ellipsis truncation, flag-off no-op, pure-context skip. `TestParseArgs_FrameLabels` in `app/config_test.go` (flag/env/config). `make test` green; gofmt + `go vet` clean.

### DIFF
--- a/.claude-plugin/skills/revdiff/references/config.md
+++ b/.claude-plugin/skills/revdiff/references/config.md
@@ -32,6 +32,7 @@ Then uncomment and edit the values you want to change.
 | `--line-numbers` | `REVDIFF_LINE_NUMBERS` | Show line numbers in diff gutter | `false` |
 | `--blame` | `REVDIFF_BLAME` | Show blame gutter | `false` |
 | `--word-diff` | `REVDIFF_WORD_DIFF` | Highlight intra-line word-level changes in paired add/remove lines | `false` |
+| `--frame-labels` | `REVDIFF_FRAME_LABELS` | Show off-screen change counts in the diff frame | `false` |
 | `--annotation-marker` | `REVDIFF_ANNOTATION_MARKER` | Prefix shown before annotation lines | `💬` |
 | `--exit-code-on-annotations` | `REVDIFF_EXIT_CODE_ON_ANNOTATIONS` | Exit 10 when annotations are produced | `false` |
 | `--no-confirm-discard` | `REVDIFF_NO_CONFIRM_DISCARD` | Skip confirmation when discarding annotations with Q | `false` |
--- a/README.md
+++ b/README.md
@@ -342,6 +342,7 @@ Positional arguments support several forms:
 | `--line-numbers` | Show line numbers in diff gutter, env: `REVDIFF_LINE_NUMBERS` | `false` |
 | `--blame` | Show blame gutter, env: `REVDIFF_BLAME` | `false` |
 | `--word-diff` | Highlight intra-line word-level changes in paired add/remove lines, env: `REVDIFF_WORD_DIFF` | `false` |
+| `--frame-labels` | Show off-screen change counts in the diff frame, env: `REVDIFF_FRAME_LABELS` | `false` |
 | `--annotation-marker` | Prefix shown before annotation lines, env: `REVDIFF_ANNOTATION_MARKER` | `💬` |
 | `--exit-code-on-annotations` | Exit 10 when annotations are produced, env: `REVDIFF_EXIT_CODE_ON_ANNOTATIONS`, config: `exit-code-on-annotations` | `false` |
 | `--no-confirm-discard` | Skip confirmation when discarding annotations with Q, env: `REVDIFF_NO_CONFIRM_DISCARD` | `false` |
--- a/app/config.go
+++ b/app/config.go
@@ -38,6 +38,7 @@ type options struct {
 	AnnotationMarker      string   `long:"annotation-marker" ini-name:"annotation-marker" env:"REVDIFF_ANNOTATION_MARKER" default:"💬" description:"prefix shown before annotation lines"`
 	ExitCodeOnAnnotations bool     `long:"exit-code-on-annotations" ini-name:"exit-code-on-annotations" env:"REVDIFF_EXIT_CODE_ON_ANNOTATIONS" description:"exit 10 when annotations are produced"`
 	VimMotion             bool     `long:"vim-motion" ini-name:"vim-motion" env:"REVDIFF_VIM_MOTION" description:"enable vim-style motion preset (counts, gg, G, zz/zt/zb, ZZ/ZQ)"`
+	FrameLabels           bool     `long:"frame-labels" ini-name:"frame-labels" env:"REVDIFF_FRAME_LABELS" description:"show off-screen change counts in the diff frame"`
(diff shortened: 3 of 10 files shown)


umputun:
going to pass on this one, and I'd rather give you the real reasoning than a one-liner.

the main thing: revdiff already answers "how much change is around me", and it does it in hunks, which is the unit the whole tool works in. The status bar shows `hunk X/Y` when the cursor is on a change and `N hunks` otherwise (`app/ui/view.go:318`), and `[`/`]` jump hunk to hunk. So on `hunk 2/5` you already know there's 1 hunk above and 3 below. `--frame-labels` restates that as `+x/-y modified lines`, a finer unit that doesn't match how you navigate, collapse, or read position everywhere else. That lines-vs-hunks split is the part that reads as off to me.

the phrasing (`no modified lines above`) and hanging a counter on both the top and bottom border also feel heavy for a pane that already carries a scrollbar and the header. That part is taste, but it pushed me the same way.

setting aside whether it should exist, it's not mergeable as-is:

- collapsed mode counts wrong. With `--frame-labels --collapsed` (or the `C` toggle), the counts fold in collapsed-hidden removed lines. A delete-only file scrolled to the end shows removals "below" instead of "no modified lines below", and a hunk that starts with hidden removals shows removals "above" while you're at the top. `modifiedAbove`/`modifiedBelow` slice the raw `m.file.lines` and run `CountChanges` over it, which has no idea which rows are hidden.
- it recomputes on every render frame. `applyDiffTopLabel`/`applyDiffBottomLabel` run inside `View()`, each doing `visualRowToDiffLine` plus a full-file `CountChanges`. That puts O(N) work back on the wheel-scroll path that #179 kept O(1) per event. Off by default, so it only hits people who enable it, but for them it undoes that fix.
- lint is red on the branch, 7 `intrange`/`minmax` issues in the new files.
- `renderFrameLabel`, `renderFrameSpans`, and `cutRightToWidth` are standalone funcs called only from `Model` methods, so per the codebase convention they should be methods (`truncateLeftToWidth` is the same shape and is one). And `cutRightToWidth` duplicates `runewidth.Truncate(s, w, "")`, which is already used in `diffview.go:90`.

if you want to come back to the idea, the shape I could see fitting is hunk-based rather than line-based, and it'd have to clear a real gap over the existing `hunk X/Y` indicator. Right now I don't think it does.

fkirill:
First things first. 
When we talk about changes we never talk about hunks, but always talk about lines. And that is for a reason. Lines carry more meaning, while hunk is just an implementation detail (historically coming from git internals). And while I undertand that revdiff necessarily deals with hunks internally, that doesn't mean that this is the most useful representation. Hunks can be useful in some scenarios, but I would argue that this is an auxiliary piece of data

If we disagree on that, then I kindly ask you to skip the rest.

Well, I guess performance impact can be easily dealt with (just precompute the numbers for every vertical screen offset, this is static info per file, and voila, back to O(1), could recompute on each display mode change). Doable.

When it comes to visual assessments (i.e. 'looks overloaded', etc.), that's the matter of judgement. I'm open to discussion about the representation. I'm sure sure we can find satisfactory solution in this area.

The representation I suggested also carries powerful visual cues. When I have to decide whether to scroll further down in the current file, I don't even have to read what the indicator says: basically if it contains color (+20 is green, -30 is red) then there are changes, keep scrolling, if it is the same color as the border, then it doesn't. You can determine if there are colored characters at the frame border extremely quickly. 
For hunks that wouldn't work, I would need to read both numbers 5/7 and analyze them. Also when you are in the hunk and out of hunk the content is at a different position and has a different format, so first you have to identify where the information is and then you have to read it, and then you analyze it. This is a way more complex mental process compared to the above.

In short, color markers and static placement is as important as the actual content it shows.

chris-peterson:
@fkirill you may want to consider this [ai-assisted diff tool](https://chris-peterson.github.io/moor/#/) (shares lineage with `kdiff3`, so is more in line with traditional diff tools)

that said, I'm actively experimenting in using `revdiff` _instead_ as I think it will reduce some of the friction that I'm experiencing with my [Claude plugin suite](https://chris-peterson.github.io/claude-marketplace/#/); particularly, the transition across "checkpoints" whether that be commits, or PR/MRs.  if the agent has done a good job, then the human pass is more of a sniff test than a line-by-line sign off

either way, thank you both for providing different perspectives as we navigate this crazy way of building software 😄

umputun:
i do work in hunks, and the git UIs I like lean the same way, so that part isn't going to move for me.

where I'll give ground: the current implementation is too heavy for what it delivers, a counter hung on both borders in `+x/-y modified lines` phrasing. That's more UI than the payoff justifies.

there may be room for a lightweight "changes above / below the viewport" cue though. I don't need it myself, so I won't pretend to spec the exact UI, but the bar is: minimal enough that it doesn't get in the way of someone who doesn't want it, and still useful for someone who does. The natural home for that is the status line, next to where `hunk X/Y` already lives, not the frame border. Something compact there could work.

if you want to explore that shape, I'm open to looking.

umputun:
@chris-peterson this reads as self-promotion more than something that adds to the thread, two links to your own projects on a closed PR about a specific feature. Please don't use my PRs for that. There are better ways to get your work in front of people.

fkirill:
Right, okay, no worries. Thanks for your time!

PR #268, opened 2026-07-16
feat(plugins): add automatic Codex plan review

### LINKED ISSUE #267: Add opt-in automatic Codex plan review via Stop hook
## Problem

`revdiff-planning` automatically reviews Claude Code plans through `PreToolUse/ExitPlanMode`, but Codex users only have the manual `revdiff-plan` skill. The Codex README still says Codex lacks hook support, and the manual skill uses best-effort rollout discovery that is ambiguous when several Codex sessions run concurrently.

Codex CLI now supports lifecycle hooks bundled in plugins. Live testing on Codex CLI 0.144.4 confirms that a root `Stop` hook can return `{"decision":"block","reason":"..."}` to continue the turn with feedback. Two details differ from the initial design assumptions:

- `permission_mode` is the approval policy (`bypassPermissions` in the tested Plan-mode session), not the collaboration mode, so it cannot gate Plan mode;
- `last_assistant_message` is present but `null` in the tested `Stop` payload. The same payload provides `transcript_path`, `session_id`, and `turn_id`.

## Proposed solution

Extend the existing opt-in `plugins/revdiff-planning` package without changing the revdiff binary or replacing the manual fallback:

- keep the current Claude `PreToolUse/ExitPlanMode` hook unchanged;
- add a Codex plugin manifest and Codex-specific hook config using `Stop`;
- review only complete `<proposed_plan>` blocks from root `Stop` events; clarification and normal Build-mode responses have no block and are ignored;
- prefer a string `last_assistant_message`; when Codex sends it as `null`, read only the exact `transcript_path` whose filename contains the event `session_id`, then select the final assistant message matching the event `turn_id`;
- do not skip `stop_hook_active == true`, because that is the revised-plan pass after annotation feedback;
- open the whole plan in revdiff through the existing trusted launcher chain;
- on a clean review, return `{}` and let the turn end;
- on annotations, return `decision: "block"` with the annotations as the reason so Codex revises the plan and the next `Stop` reviews the new revision;
- fail open with a visible warning for malformed or mismatched input, missing revdiff, missing launcher, or overlay failure, so the session cannot be wedged;
- document Codex marketplace install plus hook trust through `/hooks`, while keeping `/revdiff-plan` as the manual fallback and correcting the obsolete no-hook statement.

This remains opt-in because it is packaged as the separate `revdiff-planning` plugin.

## Test plan

Add integration coverage for Claude and Codex payloads: clarification skips, clean plans, bypass approval, annotations and revised-plan loops, null-message exact-turn transcript fallback, missing message/binary, launcher failure, and malformed JSON. Verify the fallback ignores other turns and rejects mismatched sessions. Keep manual skill behavior intact.

Live smoke coverage on Codex 0.144.4 should prove: clean plan opens and quits normally; an annotation blocks the turn, causes a full revision, and opens RevDiff again; Build-mode/non-interactive output is unaffected.

If this direction fits the project, I can send a focused PR with no binary/API changes, changelog edits, or version bump.

stevensuna:
## Problem

`revdiff-planning` automatically reviews Claude Code plans through `PreToolUse/ExitPlanMode`, while Codex users only have the manual `revdiff-plan` skill. The documentation also still says Codex has no hook support.

Codex plugin hooks now expose a `Stop` event, but its runtime contract differs from Claude's: Codex 0.144.4 reports `last_assistant_message` as `null` in the tested Plan-mode session and supplies the exact `transcript_path`, `session_id`, and `turn_id` instead.

Closes #267.

## Solution

- Add an opt-in Codex manifest and `Stop` hook to the existing `revdiff-planning` plugin.
- Preserve Claude's `PreToolUse/ExitPlanMode` hook in a Claude-specific hook file.
- Review only complete `<proposed_plan>` blocks and skip clarification or normal Build-mode turns.
- Prefer `last_assistant_message`; when it is null, read only the event's matching transcript, session, and turn.
- Return `{}` after a clean review and `decision: "block"` with annotations when Codex must revise the full plan.
- Fail open with a visible warning for malformed input, missing dependencies, transcript mismatches, and launcher failures.
- Keep `/revdiff-plan` as the manual fallback and update installation/documentation accordingly.

No revdiff binary/API behavior, changelog, or release version is changed.

## Why `Stop` instead of `ExitPlanMode`

Claude exposes the proposed plan before `ExitPlanMode` through `PreToolUse`. Codex does not use that tool event; its plugin hook receives the completed assistant turn through `Stop`. Returning a block decision continues the same Codex turn with review feedback, which creates the revise-and-review loop.

## Verification

- `make test` with Go 1.26.4 and Bun 1.3.13: pass, race detector enabled, 90.4% statement coverage
- `make lint`: pass, 0 issues
- `make build`: pass
- Python compile, JSON validation, `bash -n`, and ShellCheck: pass
- `git diff --check`: pass
- Hook integration coverage includes clarification skip, clean plan, bypass approval, annotations, revised-plan loop, exact-turn transcript fallback, session mismatch, missing message/binary, launcher failure, and malformed JSON

`shfmt -d` still reports the repository's existing space indentation in the bundled launcher scripts; this PR does not modify those scripts.


umputun (changes requested):
the Codex-parity idea is fine, but the approach needs work before it can go in, and one item is a design call on my side.

the blocker first, because it makes the feature silently do nothing in the common case. The hook assumes `last_assistant_message` carries the `<proposed_plan>` block, but Codex strips that block out of `last_assistant_message` in plan mode. So whenever the turn has any prose around the plan, the field is non-null *without* the plan, `extract_plan()` returns `None`, and the hook returns `{}` without ever opening revdiff. The hook only reads the transcript when the message is `null`, so the transcript fallback never runs for the mixed-text case. The tests pass only because every fixture makes the plan the entire message, so stripping leaves it null and the transcript path runs. Fix: consult the transcript whenever no plan is found in the message, not just when the message is null.

second, the hooks wiring, and this is the design call. I don't want the Codex hook routed through the same `hooks/hooks.json` that Claude discovers by default. The PR moves the working Claude config to `hooks/claude-hooks.json` and adds `"hooks": "./hooks/claude-hooks.json"` to the Claude manifest just so Codex can take the default filename. That makes Claude's loading depend on the explicit key suppressing default `hooks.json` discovery, which the plugin docs don't actually pin down, and I won't rely on that for an auto-firing hook. Leave the Claude hook where it is, untouched, and give Codex its own declared hooks file, or split the Codex side into a separate plugin. The two runtimes should not share the default hooks filename.

rest of it:

- no plan-mode gate. The hook fires on any Stop turn whose message contains `<proposed_plan>` tags, with no `permission_mode` check, so a normal Build/default answer that quotes those tags will launch a review. Add a mode gate.
- `phase == "final_answer"` in `message_from_transcript()` is provider-dependent. Providers that omit `phase` never match, so the fallback goes dark on them. Keying on `turn_id` plus the last assistant message is safer.
- no compare view in the revise loop. The Claude hook chains revisions through the `MARKER_RE` marker and drives `--compare-old`/`--compare-new`, so each round shows only what changed. The Codex block reason just asks for the whole plan again, so every revise round re-opens the full plan. Worth matching if the loop is meant to be usable on long plans.
- CLAUDE.md still says the launcher override chain is Claude-only and codex ignores it, but this hook uses `resolve-launcher.sh` with its own `PLUGIN_DATA` user layer, so that bullet is now wrong.

also this needs a rebase, it conflicts with master (#273 touched the same launcher dir). Please sort that out before another look.

stevensuna:
Thanks for the detailed review. I pushed `79fd7e2`, rebased onto the current master while keeping the agterm changes from #273.

The main bug was in the transcript fallback. It now runs whenever `last_assistant_message` does not contain a complete `<proposed_plan>`, rather than only when the field is null. The fallback selects the last assistant message from the exact session and turn without depending on `phase`. A readable turn without a complete plan is treated as a clarification; missing or mismatched transcript data warns and fails open.

I also restored the Claude hook wiring to its original shape. `hooks/hooks.json` remains the default Claude `PreToolUse/ExitPlanMode` config, while the Codex manifest explicitly loads `hooks/codex-hooks.json`.

The Stop hook is now limited to Plan mode, and annotated revisions use the same guarded snapshot chain as Claude for rolling old/new comparisons. The launcher documentation now covers Codex’s `PLUGIN_DATA` override as well.

The targeted tests and the full test, lint, and build commands pass locally. The Actions run has not started because GitHub requires maintainer approval for workflows from the fork: [run 29568577038](https://github.com/umputun/revdiff/actions/runs/29568577038).

umputun (commented):
all 7 points from the last round are addressed - transcript fallback now fires on any planless `last_assistant_message`, the hooks split keeps Claude's default `hooks.json` untouched with Codex loading `codex-hooks.json`, the plan-mode gate is in, `turn_id` keying replaced `phase`, the revise loop drives `--compare-old/--compare-new`, and the CLAUDE.md override bullet is fixed. The tests exercise the real hook, not mocks. A few new things from this pass, none blocking on correctness:

**worth fixing this round**

1. `codex-plan-review-hook.py:231-244` - the block reason drops the anti-substitution warning the Claude hook carries (`plan-review-hook.py:237-241`). Without it a drifting agent can reuse an older `plan-rev-*` marker and compare against the wrong baseline, and since both hooks share `$TMPDIR` and the `plan-rev-` prefix and you `unlink` the old snapshot on success, that can delete a concurrent review's snapshot. Mirror the Claude hook's "do NOT substitute any other plan-rev-*.md path" sentence.

2. `codex-plan-review-hook.py:79-115` - `message_from_transcript` keeps the turn's *last* assistant message, then `main()` silently skips if it has no plan. If Codex emits the plan followed by a planless closer in the same turn, the plan is dropped and the review never opens, the same silent no-op class as the round-1 blocker. Track the last message that matches `PLAN_RE`, fall back to last-overall only if none has a plan. Add a plan-then-planless test.

3. `codex-plan-review-hook.py:188-194` - `new_snapshot = Path(tmp.name)` is bound after `tmp.write(plan)`, so a write failure orphans the temp file (the `except OSError` skips the unlink and logs "launcher could not be started", which misattributes it). The `KeyboardInterrupt` handler also never unlinks. Bind the path before the write, and unlink in the interrupt handler.

4. `app/plugin_exit_code_test.go` - `trusted_snapshot()` only has positive coverage. Add a rolling-review case with a marker pointing outside `$TMPDIR` or without the `plan-rev-` prefix, asserting a single-arg launch (`argsLog "1\n"`) and the marker line stripped from the new snapshot. It's the one guard between an injected marker and an arbitrary `--compare-old` file.

5. `CLAUDE.md:64` - this adds a third versioned manifest (`.codex-plugin/plugin.json`) but the release rule still says to bump only `plugin.json` + `marketplace.json`. Update it to name the codex manifest so the next release doesn't leave it stale.

**the one that matters most, please confirm before merge**

the transcript fallback is the common path (Codex strips the plan out of `last_assistant_message`), and it rides on two unverified assumptions about the rollout format: the filename ends `-{session_id}.jsonl` and each item carries `internal_chat_message_metadata_passthrough.turn_id`. Every test synthesizes that shape, so if it doesn't match real Codex output the feature silently no-ops while CI stays green. Can you verify both against a live Codex rollout and commit a sanitized captured fixture, so a format change fails a test instead of failing users?

**optional / your call**

- `codex-plan-review-hook.py:50-62` - the sibling hook documents *why* callers must use the resolved path (symlink TOCTOU) and why the prefix confinement matters. The copy reduces both to one-liners. Worth porting the two rationale comments on a security guard.
- snapshot trust isn't bound to the session/chain (same as the Claude hook). Folding `session_id` into the snapshot prefix would close the cross-chain compare/delete window, but it's low-frequency and throwaway data, so fine to defer.
- on older Codex builds missing the transcript identifiers, every plan-mode clarification turn warns "plan review skipped", which reads as broken. Only matters if those builds exist.

separately, `plugins/codex/skills/revdiff/SKILL.md:25` still says the override chain is "Claude-only", accurate for the diff-review launcher, but now that the planning hook honors `PLUGIN_DATA` under Codex it's ambiguous. Not in this PR's diff, just noting.

stevensuna:
I pushed `c8d3521` with the follow-up fixes from this review.

Transcript fallback now keeps the last assistant message in the exact turn that contains a complete `<proposed_plan>`, so a plan followed by a planless closer no longer becomes a silent no-op. The revision response also carries the anti-substitution warning for `plan-rev-*` markers, and snapshot creation now binds the path before writing and cleans it up on write failure, timeout, or interruption. The rolling-review regression also verifies that an untrusted marker is stripped while the launcher stays in single-argument mode.

I verified the rollout assumptions against live Codex 0.144.4 and 0.144.5 data: the rollout filename ends in `-{session_id}.jsonl`, and assistant response items carry `internal_chat_message_metadata_passthrough.turn_id`. The commit includes a sanitized captured 0.144.4 rollout fixture and exercises the real fallback against its original envelope and filename shape. I also updated the release guidance to keep the Claude manifest, Codex manifest, and marketplace entry synchronized.

The full race-enabled test suite passes locally with 90.4% coverage; lint reports 0 issues, and the build plus targeted hook, Python, JSON, and whitespace checks pass. GitHub has not attached checks to the new head yet. The PR is mergeable and ready for another look.

umputun (commented):
the round-2 items are all in - anti-substitution warning, plan-message selection, snapshot ordering + interrupt cleanup, the untrusted-marker rejection test, the CLAUDE.md sync rule, and the live-rollout fixture that pins the `-{session_id}.jsonl` + `internal_chat_message_metadata_passthrough.turn_id` format. Tests pass. One edge left in the transcript selection:

`codex-plan-review-hook.py:109-118` - `PLAN_RE.search` treats an empty `<proposed_plan></proposed_plan>` as plan-bearing and sets `plan_message`, but `extract_plan` returns None on an empty block. So a real plan followed by an empty block (a later message that echoes the template, or a trailing empty block in the same message) drops back to a silent no-op, the same class the `plan_message` change was meant to close, just narrower. Confirmed against the actual regex: `[valid plan, "...<proposed_plan></proposed_plan>"]` extracts None, while `[valid plan, planless closer]` works.

fix: have `extract_plan` return the last *non-empty* complete block, and track `plan_message` with the same non-empty test (`if extract_plan(message): plan_message = message`). Add regressions for a valid plan followed by an empty-block closer, and a valid-plus-empty block in one message.

two nits:
- `CLAUDE.md:64` - "the root `marketplace.json`" should be `.claude-plugin/marketplace.json`, there's no marketplace.json at the repo root.
- the untrusted-marker test covers the non-`plan-rev-` prefix branch but not the outside-`$TMPDIR` branch. A marker resolving outside tempdir would round out `trusted_snapshot`'s coverage.

stevensuna:
Fixed in `5050afc`.

`extract_plan()` now scans complete blocks from the end and returns the last non-empty plan. Transcript selection uses the same extraction result, so a later empty-block message cannot hide an earlier valid plan, and a valid-plus-empty pair in one message still extracts the valid block.

I added regressions for both empty-block cases, plus the canonical-path-outside-`TMPDIR` branch of `trusted_snapshot()`. I also corrected the release note to point to `.claude-plugin/marketplace.json`.

The focused hook test and the full `go test ./...` suite pass locally with the pinned Go/Bun toolchains, and `git diff --check` is clean.

umputun (approved):
all four rounds resolved. The empty-block edge is closed - `extract_plan` returns the last non-empty block and selection uses the same non-empty test, with regressions for both empty-block orderings and the outside-`$TMPDIR` marker branch.

lgtm.

PR #272, opened 2026-07-16
Feature/agent cmd

JalenEvans:
## Summary

Adds `--agent-cmd` / `REVDIFF_AGENT_CMD`, which pipes the current annotation output to an external command's stdin on the `O` in-session flush (`flush_output` action). This lets revdiff hand annotations straight to an agent/relay process on each flush, instead of requiring an external file-watcher on the `-o` output.

The new sink composes with `-o/--output`: a single `O` flush can target the output file, the agent command, or both — the flush is gated on either sink being configured, and each send carries the full current annotation snapshot (`FormatOutput()`), not a delta.

## Key changes

- New `app/agent` package: `Runner.Send` runs `--agent-cmd` through the system shell (`sh -c` / `cmd /c` on Windows) with a 30s timeout, writes the payload to stdin, and folds a bounded slice of stderr into the returned error on failure or timeout.
- `app/ui`: consumer-side `AgentSink` interface (`Send(payload string) error`), wired via `ModelConfig.Agent` with the same typed-nil-collapsing guard used for `Editor` (nil disables the subsystem). The send runs asynchronously via a `tea.Cmd` (`agentFlushedMsg`) so a slow or hung command never blocks the TUI.
- `handleFlushOutput` reworked to attempt the file write (if `-o` is set) and/or kick off the agent send (if `--agent-cmd` is set), with distinct in-flight/success/failure status-bar hints ("Sending…", "Sent N annotations to agent", "Agent flush failed", etc.).
- `app/config.go` / `app/main.go`: new `AgentCmd` option, wired into a `ui.AgentSink` at the composition root.
- Docs synced: README, `site/docs.html`, plugin `config.md`/`usage.md` references for both Claude and Codex, all three plugin `SKILL.md` files (noting the flag is standalone-only — not part of the managed plugin launch flows), and an `Unreleased` CHANGELOG entry.
- Tests: new `app/agent` unit tests (success, non-zero exit, timeout, stderr truncation) and extended `app/ui/output_test.go` covering the file-only, agent-only, both-sinks, and neither-sink flush paths.

## Testing

- `make build`, `make test` (race detector + coverage), and `make lint` all pass locally.

## Notes / risks

- The agent command is run through the system shell by design (`--agent-cmd` / `REVDIFF_AGENT_CMD` is user-supplied, same trust model as `$EDITOR`) — flagged with a `//nolint:gosec` at the call site.
- Snapshot semantics only: each flush re-sends the full current annotation set, matching the existing `-o` flush behavior; there's no diffing or append mode.


JalenEvans:
This also adds the ability to integrate with Coding Agents that can push into a live session using the stdin, like opencode. When/If merged, a follow-up opencode integration for that will be opened.

JalenEvans:
I like to have revdiff open as a pane in cmux and always run as a diff. Right now, I have a work around where I write annotations to a file that are being watched to the background and when that file is written to it uses cmux send-key to send the prompt. This allows for people to make simple scripts to integrate with tools like cmux or herdr to accomplish this.

Hopefully in the future, more agents will allow for live sessions to be pushed to with the stdin.

umputun:
had a good look at this. The implementation is clean, well tested, and follows the existing patterns (dedicated `app/agent` package, consumer-side interface, typed-nil collapse like `Editor`). No complaints on the code itself. My hesitation is about scope, not quality.

the `O` flush to `-o` was built for exactly this agent handoff loop: annotate, flush, let an agent pick up the annotations, `R` reload, continue (see `docs/plans/completed/20260702-in-session-annotation-flush.md`). `--agent-cmd` doesn't add a new capability, it removes one hop, the watcher that relays the output file into the agent, which is your current setup. So the workflow already works today with a small external script.

I did check whether `-o >(relay)` process substitution would cover it with zero core change, and it won't: `AtomicWriteFile` does temp-file + rename, so a `/dev/fd/..` sink breaks. So the watcher really is the current path, fair enough.

the cost side is what gives me pause. This is a new package, a new interface, an async message path, and six doc surfaces, plus a carve-out in all three managed plugin flows (Claude, codex, pi) that have to say "do not pass `--agent-cmd`". That is a fair bit of core and doc surface to carry for something that already works via a small watcher, and it is aimed at a narrow slice: people running revdiff standalone who want push-to-agent and won't wire that watcher.

two things that would move me:
- demand beyond your own setup. If this is mostly your cmux/herdr workflow, a documented watcher snippet in the README might serve the same need with no core code.
- the shell-exec divergence. `app/editor` deliberately stays shell-free (argv split, no shell) while `app/agent` runs everything through `sh -c`. If this goes into core, I would want that reconciled or justified.

so my lean right now is to keep this out of core, but it is a soft no, not a hard one. Make the case on those two points and I will happily reconsider.

JalenEvans:
Thank you for the amazing feedback. Your in-depth documentation really helped the process. I truly appreciate the speed and quality of your review and have been thoroughly enjoying using revdiff. Now I will respond to your concerns. 

> "demand beyond your own setup."

I understand you concern. However, I think that these tools (cmux, herdr, tmux, and other terminal multiplexers) will become more popular as more Software Engineers use more agentic workflows. I think having different options for different workflows would benefit this change. However, I am ultimately submitting this for my own benefit.

I have made the opencode integration that builds off of this (which I will rebase on top of these changes if merged) and it takes the setup from 30 minutes of trial and error and downloading dependencies to instant integration. I understand that it only works for one coding agent but opencode is also growing at a great rate. If you would like to test it out I can push it on my fork and let you know when it is fully flushed out so you can test it yourself.

> "the shell-exec divergence."

Will be making those changes and updating this pull request. 

Thank you again.

umputun:
the shell-exec turnaround is what I wanted there, argv-split and shell-free, and it lines up with how `app/editor` already works. That said, it doesn't change my call on scope.

on scope I'm passing on `--agent-cmd` in core. The refactor answered how it's built, not why it belongs in the binary. The annotate, flush, hand-to-agent loop already works today through a small watcher on the `-o` file, which is your current setup, so `--agent-cmd` removes one hop for the standalone-with-no-watcher case rather than adding a capability. Against that it's a new package, a new interface, an async path, and more doc surface, and the demand is still your own workflow. Not enough to carry in core.

best home for this is your fork, paired with the opencode integration you built on top. If a documented watcher snippet in the README would help others hit the same loop without the extra script, I'm happy to take that as a docs PR.

the call is purely scope, the implementation is solid. Appreciate the fast turnaround on the shell fix and the back-and-forth.

unrelated, just fyi, since you mentioned cmux/herdr/tmux and agentic terminal workflows growing, I maintain one in that space too, https://github.com/umputun/agterm, might be up your alley.

JalenEvans:
Thank you for your reviewing. I think I might abstract the logic and make a tool that sends annotations from a file into a coding agent session and then use that with revdiff -o. I'll submit a README edit with the tool under a new Related Projects section if that is okay.

I did look at agterm! I run on Linux so I might try out the linux fork. Awesome projects!

PR #276, opened 2026-07-18
fix(ui): restore mouse after source editor return

### LINKED ISSUE #275: Mouse tracking is not restored after source editor exits
When opening a source file with e and returning from the editor, mouse-wheel input can behave incorrectly because Bubble Tea v1 does not restore mouse tracking after tea.ExecProcess.

 There seem to be two possible fixes:

 1. Fix it upstream in Bubble Tea. This was previously proposed in https://github.com/charmbracelet/bubbletea/pull/1583, but the PR was closed after Bubble Tea’s renderer architecture changed.
 2. Add a small revdiff-local workaround: mirror the existing annotation-editor behavior and return tea.EnableMouseCellMotion after the source editor exits, while preserving reload behavior and respecting --no-mouse.

 I have the second approach implemented and tested locally. If that approach is acceptable, I’d be happy to follow up with a PR.

Really liking revdiff btw. TY

phewitt:
## Problem

Opening a source file with `e` releases the terminal through `tea.ExecProcess`, but the source-editor completion path does not restore mouse tracking afterward. This leaves mouse input disabled after both clean and error returns.

## Solution

Mirror the existing annotation-editor behavior by re-enabling mouse cell motion when the session started with mouse tracking enabled. Preserve the source-editor reload behavior by batching restoration with the reload command when both are needed, while continuing to respect `--no-mouse`.

Fixes #275

### DIFF
--- a/app/ui/editor.go
+++ b/app/ui/editor.go
@@ -113,6 +113,7 @@ type sourceEditorFinishedMsg struct {
 	err                  error
 	fileName             string
 	reloadAfterCleanExit bool
+	restoreMouse         bool
 }
 
 // sourceEditorTargetResult is the UI-side decision for one source-editor
@@ -153,7 +154,12 @@ func (m *Model) openSourceEditor() tea.Cmd {
 		}
 	}
 	return tea.ExecProcess(cmd, func(runErr error) tea.Msg {
-		return sourceEditorFinishedMsg{err: runErr, fileName: result.fileName, reloadAfterCleanExit: result.reloadAfterCleanExit}
+		return sourceEditorFinishedMsg{
+			err:                  runErr,
+			fileName:             result.fileName,
+			reloadAfterCleanExit: result.reloadAfterCleanExit,
+			restoreMouse:         m.cfg.mouseTracking,
+		}
 	})
 }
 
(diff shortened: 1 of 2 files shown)


umputun (approved):
logic looks correct and matches the annotation path from #235. All five return paths restore mouse cell motion, `--no-mouse` still wins via the `mouseTracking` gate, and batching with the reload cmd is safe. One test-coverage nit before merge, the added tests don't actually pin the fix in place:

- `openSourceEditor` sets `restoreMouse: m.cfg.mouseTracking` (`app/ui/editor.go:154`) inside the `tea.ExecProcess` completion closure, but no test invokes that closure, so deleting the line leaves every test green while reintroducing the bug. A case that sets `mouseTracking=true` and asserts the propagated flag would catch it.
- the stale-tree (`:359`) and changed-file (`:363`) branches return `restoreMouseCmd`, but the only tests covering them (`SkipsReloadWhen*`) leave `restoreMouse` false and assert `cmd == nil`, so reverting either to `return m, nil` still passes. Extend the `RestoresMouseTracking` table with a `restoreMouse:true` case that hits each mismatch branch.

both minor, production behavior is right on all five paths.

PR #285, opened 2026-07-22
Add filterable file jump picker

### LINKED ISSUE #284: Add a quick file picker for large diffs
When I’m reviewing a diff with a lot of files, I often know which file I want but still have to step through the sidebar to reach it.

I’d like to add a configurable `jump_file` action, bound to `Ctrl+P` by default. It would open a small picker where typing filters the visible file paths, the normal up/down keys move through matches, and Enter loads the selected file.

My intention is to keep the patch small by following the existing theme-picker interaction and reusing the current file-tree selection and loading paths. Active annotated-only or unreviewed-only filters would still be respected.

I have a working implementation on my fork for personal use. If this fits the project direction, I’d be happy to submit it as a focused PR.

kalifg:
## Summary

- add a configurable `jump_file` action with `Ctrl+P` as the default
- filter the current sidebar’s visible paths with case-insensitive substring matching, preserving annotated-only and unreviewed-only filters
- reuse `SelectByPath` and `loadSelectedIfChanged` when jumping, while clearing pending annotation and hunk navigation
- preserve search-modal precedence so `Ctrl+P` continues to recall history while the `/` prompt is active
- cover no-match behavior, both tree filters, keyboard and mouse selection, scrolling, resizing, truncation, and modal key precedence

The picker follows the existing overlay pattern and adds no dependency or generic picker/fuzzy-search framework.

## Testing

- `make test`
- `gofmt` check
- `golangci-lint v2.11.4 run`
- `git diff --check`

Closes #284

Implemented with Codex assistance; I reviewed the diff and tested the behavior locally.


umputun (commented):
scope matches what we agreed on in #284, so I don't think any scope correction is needed, but three blocking issues need to be fixed before this can merge:

1. `app/ui/overlay/filepicker.go:168` handles `ActionDown`/`ActionUp` before `tea.KeyRunes`, and with the default keymap `j` and `k` resolve to those actions in `Model.handleModalKey`, so they move the cursor instead of entering the filter and make any query containing either letter impossible to type. Pls give unmodified printable runes filter-input priority, keep navigation actions for non-printable or modified keys, and add a model-level regression that types filters containing both `j` and `k`.

2. `app/ui/filepicker.go:22` can leave an older async load valid when the picker selects the file already on screen: if A is displayed, selecting B starts a load, and the picker selects A before B returns, `loadSelectedIfChanged` skips the load without advancing `file.loadSeq`, so the late B response passes the check in `app/ui/loaders.go:504` and replaces A. Track the requested path or reject a `fileLoadedMsg` when its path no longer matches the selected target, with a delayed-response regression for this sequence.

3. `app/ui/overlay/filepicker.go:116` renders repository paths without sanitizing the display copy, which lets filenames containing newlines, ANSI/OSC sequences, C0/C1 bytes, or bidi controls add visual rows, emit terminal controls, and break the fixed-row mouse mapping. The sanitizer in `app/ui/view.go:120` already handles this input, so pls reuse or move it for picker display while keeping the raw path for filtering and selection, then cover newline, escape/control bytes, invalid UTF-8, and bidi controls with an assertion that each entry stays on one visual row.

smaller follow-ups:

- Copilot's two comments about the Esc wording are correct and still open; in both mirrored usage docs, say that the first `Esc` clears a non-empty filter and keeps the picker open, while the second closes it.
- update `CLAUDE.md:22` to include the file picker in the `app/ui/overlay/` package description.
- add the missing `ActionJumpFile` pending-chord cleanup test in `app/ui/model_test.go`, together with Shift+wheel half-page movement and boundary tests in `app/ui/overlay/filepicker_test.go`.
- make `truncateFilePath` a method on `filePickerOverlay`, matching the project convention in `CLAUDE.md:35` and the neighboring file-tree helper.

umputun (commented):
three things block merge, the rest is minor.

round 1's blockers are closed. I re-traced each one rather than trusting the commit message: printable runes take priority now, the `requestedPath` / `canceledLoadSeq` pair drops the late response across A→B→A, A→B→C→A and reload-mid-flight, and `formatEntry` sanitizes the display copy while filtering and selection keep the raw path.

**blocking**

`app/ui/loaders.go:379` and `app/ui/overlay/filepicker_test.go:168` - golangci-lint fails on `gocritic evalOrder` and `intrange`, both on lines this PR adds, so the build job is red. Fix: `cmd := m.requestFileDiff(f); return m, cmd`, and `for range paths`.

`app/ui/overlay/filepicker.go:211` - `appendPrintableRunes` gates on `tea.KeyRunes`, but bubbletea delivers a lone space as `tea.KeySpace` (`vendor/github.com/charmbracelet/bubbletea/key_sequences.go:34`), so a typed space is dropped and no path containing one can be filtered by typing its name. A bracketed paste arrives as a single `KeyRunes` with the space inside and does work, which is why nothing caught it. Fix: accept `tea.KeySpace` and append `" "`, with a case over a path like `docs/release notes.md`.

`app/ui/overlay/filepicker.go:110` - `renderFilter` emits the query unbounded inside a box styled `.Width(f.popupWidth)` with `Padding(1, 1)`, so anything past `popupWidth-4` soft-wraps and adds a row while `handleLeftClick:274` still assumes `entriesTop = 4`. On a 34-column pane that is a 21-character filter, and past it a left-click selects the entry below the one clicked, so the wrong file opens. Fix: truncate the display copy with `runewidth.Truncate` and keep the raw `f.filter` for matching, or have `render()` record the real pre-entry row count and read that instead of the constant.

**minor**

`README.md:730`, `site/docs.html:531`, both `usage.md` copies and `docs/ARCHITECTURE.md:165` - the picker paragraph is wrong in three ways. "the configured up/down keys" do not navigate, since the shipped `j` / `k` bindings type into the filter and only the arrows move; `Esc` "clears a non-empty filter before closing" is still the pre-fix wording in the README and docs.html, though the plugin copies already say it in two steps; and the `Ctrl+P` row's "(recalls previous history while the search prompt is active)" repeats the Search table's own `↑` / `Ctrl+P` row a few lines below. Fix: say the arrows move and printable keys always filter, reuse the two-step `Esc` sentence from the plugin copies, and drop the parenthetical.

`docs/ARCHITECTURE.md:150` - the `VisibleFiles()` sentence sits at column 0 between the `FileTree` and `TOC` bullets, which terminates the list in CommonMark, so the section renders as a one-item list, a stray paragraph, then a second one-item list. Fix: fold it into the `FileTree` bullet.

`app/ui/style/resolver.go:41` - the sanitizer is a receiverless method on `Resolver`, so `view.go:122` and `filepicker.go:118` each build a throwaway `style.Resolver{}` to reach it and `Model.sanitizeFilenameForDisplay` is now a pass-through. The move also dropped the godoc rationale tying the strip set to the diff-header invariant that `.claude/rules/gotchas.md` names this function as the enforcer of. Fix: make it a package-level `style.SanitizeFilenameForDisplay`, matching `style.AnsiFg` and `diff.SanitizeCommitText`, and carry the original rationale onto it.

`app/ui/overlay/filepicker.go:137` - `truncateFilePath`'s fallback branch is a near-verbatim copy of `Model.truncateLeftToWidth` (`app/ui/view.go:127`), differing only in using `runewidth.StringWidth` where the original uses ANSI-aware `lipgloss.Width`. Fix: move `truncateLeftToWidth` into `app/ui/style` beside the sanitizer this PR already relocated, and keep only the `…/basename` preference here.

`app/ui/overlay/filepicker.go:18` - the picker calls its column constant `horizontalChrome` where both siblings use `horizChromeCols`, carries no row-map comment on `handleLeftClick` (`themeselect.go:295` spells the offsets out), and renders with `style.StyleKeyThemeSelectBox`. `.claude/rules/gotchas.md` cites `horizChromeCols` by name, so a grep on it finds two overlays of three, and re-padding what is documented as the theme selector's box silently breaks the picker's click mapping. Fix: rename to match, copy themeselect's row-map comment, and add a picker style key or give the shared one a neutral name.

`app/ui/filepicker_test.go:118` - `TestModel_JumpFilePreservesSidebarFilters` passes whether or not the picker respects the tree filters: `b.go` is index 1 in the filtered and the unfiltered list alike, so one `KeyDown` reaches it either way. Fix: send `KeyDown` twice and keep asserting `b.go`, which clamps there when filtered and moves to `c.go` when not, where `SelectByPath` fails and the jump silently no-ops.

umputun (approved):
no blockers this round. Round 2's three are closed, and I re-traced each rather than trusting the commit messages: both lint rules are fixed structurally and CI is green on 3bfc036, `tea.KeySpace` reaches the filter, and the truncated filter row keeps `entriesTop = 4` correct. The sanitizer and truncator cases moved to `app/ui/style/display_test.go` intact.

merging this now. Five minors below, none of them blocking - pick them up in a follow-up if you want to, otherwise I'll get to them.

**minor**

`app/ui/overlay/filepicker.go:139` and `:110` - `truncateFilePath`'s guards and `renderFilter` measure with `runewidth` while lipgloss paints grapheme widths, and the two disagree by a cell on VS16 sequences and regional-indicator flags, so a path carrying three or more of them clears the budget check, wraps, and shifts every row below it until a left-click opens the file above the one clicked.
Fix: `lipgloss.Width` for the `<= width` and candidate guards, `ansi.Truncate` in `renderFilter`, plus a case over a path with three VS16 graphemes at the exact budget width.

`app/ui/loaders.go:491` - `return m, tea.Batch(m.requestFileDiff(f), statsCmd)` is the same unspecified evaluation order as `:379` from last round, which got split while this one did not, and `gocritic evalOrder` only inspects direct call operands of the return so nothing caught it.
Fix: `cmd := m.requestFileDiff(f); return m, tea.Batch(cmd, statsCmd)`.

`app/ui/overlay/filepicker.go:184` - the picker deletes only on `tea.KeyBackspace`, which is DEL, so ASCII BS arriving as `tea.KeyCtrlH` falls through; on a terminal that sends BS the Backspace key still works in the search prompt, since `bubbles/textinput` binds both, and does nothing here.
Fix: accept `tea.KeyCtrlH` alongside `tea.KeyBackspace`, with a parity case.

`app/ui/loaders.go:374` - the godoc still describes only "loads the selected file if it changed", but the no-load branch now records the cancellation that makes the rejection at `:519` reachable, and that branch carries no comment while 13 callers share the side effect.
Fix: state the cancellation in the godoc and point the branch at `handleFileLoaded`.

`app/ui/style/display.go:12` - the move kept the header-invariant clause and the per-arm codepoint comments but dropped the paragraph saying the bidi and format strip is anti-spoofing rather than width math, and `TruncateLeftToWidth` lost the documented returns (`""` at budget <= 0, `"…"` at budget 1) that `display_test.go:51` still asserts.
Fix: restore those two sentences.

PR #286, opened 2026-07-22
fix(themes): resolve --theme auto from tmux client list instead of current client

kylesnowschwartz:
`--theme auto` inside tmux now reads themes from `tmux list-clients` across the whole server and picks the most recently active client that reports one, instead of asking `display-message` for the current client's `#{client_theme}`.

Why: when revdiff runs in a detached session that a `display-popup` client attaches to (the plugin launcher's tmux backend), tmux resolves the popup's own nested client as current. A nested tmux client never learns a theme, so `#{client_theme}` expands empty, the masked OSC 11 fallback also gets no answer in that topology, and detection lands on the dark default. Whether detection ran before or after the popup attach was a millisecond race, so auto themes flipped between correct and dark at random. A lost race also stalled startup for termenv's 5-second OSC timeout when stdout was a tty.

- `tmuxClientThemes()` runs `tmux list-clients -F '#{client_activity} #{client_theme}'` server-wide; the outer terminal's client always appears in that list, whichever client tmux considers current.
- `pickTmuxClientTheme()` picks the theme of the most recently active client that reports one and skips themeless clients (nested tmux clients, terminals without light/dark reporting).
- tmux < 3.5 (empty `#{client_theme}` format) and detached sessions with no attached clients fall through to the existing OSC 11 and COLORFGBG paths unchanged. Non-tmux terminals and GNU screen are untouched.

Verified on tmux 3.7b by reproducing the popup topology: with the popup client attached, the old binary resolved the dark theme on a light terminal and the fixed binary resolves the light theme. Both fall-through paths are covered by table tests in `app/termbg_test.go`.

### DIFF
--- a/app/termbg.go
+++ b/app/termbg.go
@@ -16,30 +16,38 @@ import (
 // multiplexer prefixes so the OSC query path stays enabled.
 const maskedTerm = "xterm-256color"
 
-// tmuxQueryTimeout bounds the tmux subprocess so a wedged tmux server (e.g.
+// tmuxQueryTimeout bounds each tmux subprocess so a wedged tmux server (e.g.
 // blocked on a run-shell hook) degrades to the next detection method instead
-// of stalling startup indefinitely. list-clients answers over the local
-// socket in single-digit milliseconds when healthy.
+// of stalling startup indefinitely. display-message and list-clients answer
+// over the local socket in single-digit milliseconds when healthy.
 const tmuxQueryTimeout = 500 * time.Millisecond
 
 // detectDarkBackground reports whether the terminal background is dark-ish.
 // termenv refuses to send OSC status-report queries when TERM starts with
 // "tmux" or "screen" (multiplexers historically swallowed them), so inside
 // tmux its detection never queries the terminal and always falls back to
-// "dark". two tmux-specific paths run first:
+// "dark". three tmux-specific paths run in order (tmux >= 3.5 for the first
+// two; client themes are fed by each outer terminal's native light/dark
+// reporting and need no tty):
 //
-//  1. `tmux list-clients` themes (tmux >= 3.5) — fed by each outer terminal's
-//     native light/dark reporting, needs no tty, and works no matter which
-//     client tmux considers current. The most recently active client that
-//     reports a theme wins.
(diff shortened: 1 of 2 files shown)


umputun (commented):
one minor note, otherwise the change is correct and the popup-topology fix is the right approach.

**server-wide recency pick can regress the multi-client case** (`app/termbg.go`, `detectDarkBackground` / `pickTmuxClientTheme`)

the old `display-message '#{client_theme}'` resolved through tmux's current-client lookup, which is scoped to revdiff's own session's client. `list-clients` with no `-t` enumerates every client on the server across all sessions and picks by `client_activity` recency, so a client attached to a *different* session can win. Concrete trigger: a light terminal and a dark terminal both attached to one tmux server, and the one you did not launch revdiff from is more recently active, so its theme wins and also blocks the OSC-11 fallback. Equal one-second-resolution timestamps then fall to arbitrary list order (the `equal activity keeps first listed` test codifies that).

minor: cosmetic and recoverable via `--theme`, and the trigger is niche. A current-client-first hybrid removes the regression at ~zero cost while keeping the popup fix intact:

```go
if dark, ok := parseTmuxClientTheme(tmuxClientTheme()); ok {   // display-message, current client
    return dark
}
if dark, ok := pickTmuxClientTheme(tmuxClientThemes()); ok {   // server-wide fallback
    return dark
}
return termenv.NewOutput(os.Stdout, termenv.WithEnvironment(tmuxEnviron{})).HasDarkBackground()
```

in the popup topology the current client is the themeless nested client, so `display-message` returns empty, falls through to the server-wide pick, and the outer client still wins. The startup race and 5s OSC stall stay gone because the fast `list-clients` runs before OSC. In a normal pane `display-message` resolves the launching session's client directly and never reaches `list-clients`, so an unrelated session cannot win. This keeps both helpers and both test suites. Your call whether the single code path is worth accepting the niche case as-is.

kylesnowschwartz:
Applied in 97bc3bc. `display-message` runs first and scopes normal-pane launches to the launching session's client; the server-wide `list-clients` scan only runs when the current client reports no theme, which is the popup topology. Re-verified both cases on tmux 3.7b: a normal pane resolves through the current client, and a popup-attached detached session falls through to the server-wide pick and gets the outer terminal's theme.

PR #287, opened 2026-07-25
perf(ui): index visual rows for page navigation

stevensuna:
## Problem

Page navigation recalculates the cursor visual row from the start of the diff after every cursor step. Each calculation also rebuilds annotation metadata and, in collapsed mode, rescans hunks. PgUp, PgDown, Ctrl-U, and Ctrl-D therefore become progressively slower deeper into large files.

On an Apple M4 with a 10,000-line diff and a 44-row terminal, the included benchmark measured:

```
before: 232-240 ms/op, 10.2 MB/op, 681k allocs/op
after:   47-48 ms/op,  4.0 MB/op,  90k allocs/op
```

## Solution

Build one visual-row offset index for each page motion and reuse one hunk list and annotation set throughout the cursor walk. Cursor lookup becomes constant-time per step while preserving wrapped lines, collapsed hunks, inline annotations, cursor position, and viewport behavior.

The benchmark remains in the test suite to make this path measurable.

## Verification

- `go test -race -timeout=100s ./...`
- `golangci-lint run` with v2.11.4
- focused page, wrap, collapsed, and viewport navigation tests
- `BenchmarkModel_PageNavigation`, repeated with allocation reporting

### DIFF
--- a/app/ui/annotate.go
+++ b/app/ui/annotate.go
@@ -581,6 +581,38 @@ func (m Model) cursorViewportYUsing(hunks []int, annotationSet map[string]bool) 
 	return y
 }
 
+// cursorVisualOffsets returns the top visual row for every diff line. Page
+// motions build this index once so each cursor step can read its visual
+// position in O(1) instead of rescanning all preceding lines.
+func (m Model) cursorVisualOffsets(hunks []int, annotationSet map[string]bool) []int {
+	offsets := make([]int, len(m.file.lines))
+	y := 0
+	if m.hasFileAnnotation() {
+		y = m.wrappedAnnotationLineCount(annotKeyFile)
+	}
+	for i := range m.file.lines {
+		offsets[i] = y
+		y += m.hunkLineHeight(i, hunks, annotationSet)
+	}
+	return offsets
+}
+
+// cursorViewportYFromOffsets returns the current cursor's visual row from a
+// pre-built line-offset index.
+func (m Model) cursorViewportYFromOffsets(offsets []int) int {
+	if m.file.name == "" || len(offsets) == 0 {
+		return max(0, m.nav.diffCursor)
+	}
+	if m.nav.diffCursor == -1 {
+		return 0
+	}
+	y := offsets[m.nav.diffCursor]
+	if m.annot.cursorOnAnnotation {
+		y += m.wrappedLineCount(m.nav.diffCursor)
+	}
+	return y
+}
+
 // cursorVisualRange returns the top and bottom viewport Y coordinates the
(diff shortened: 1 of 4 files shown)


umputun (commented):
lgtm on the change itself, the indexed path is equivalent and the speedup reproduces here. Two small things.

`app/ui/annotate.go:587-614`: `cursorVisualOffsets` + `cursorViewportYFromOffsets` re-derive what `cursorViewportYUsing` already computes (file-annotation base, `-1` handling, `cursorOnAnnotation` adjustment), and nothing pins the two implementations equal. Mutating away either the file-annotation base or the `cursorOnAnnotation` adjustment leaves the whole suite green: the base cancels in every `endY - startY` delta, so it only surfaces on a page move that starts or ends at `diffCursor == -1`, and no test combines a file-level annotation with page nav.
Fix: add a differential test asserting `cursorViewportYFromOffsets(cursorVisualOffsets(hunks, set)) == cursorViewportYUsing(hunks, set)` for every cursor index plus `-1` and both `cursorOnAnnotation` values, on a fixture with wrap on, a file-level annotation, a multi-row inline annotation and a collapsed hunk.

`app/ui/annotate.go:609`: `offsets[m.nav.diffCursor]` indexes unguarded where the `cursorViewportYUsing` it replaces was total over every cursor value. Nothing reaches it out of range on any current path, so this wants documenting rather than a runtime clamp.
Fix: state the precondition in the godoc, `offsets` must come from `cursorVisualOffsets` over the current `m.file.lines`, and `m.nav.diffCursor` must be in `[-1, len(offsets))`.

stevensuna:
Thank you for taking the time to reproduce the speedup and probe the equivalence boundary. Your mutation check exposed a real blind spot in my coverage: the benchmark proved the indexed path was faster, but the existing tests did not actually protect it from drifting away from the direct calculation.

I pushed `74b0bf1` with the differential test you suggested. It compares both calculations at every cursor index, including `-1`, with both `cursorOnAnnotation` states, using a fixture that combines wrapped content, multi-row file and inline annotations, and a collapsed hidden line. Removing either the file-annotation base or the annotation-row adjustment now makes the test fail.

I also documented the offset provenance and cursor-range preconditions on `cursorViewportYFromOffsets`. The full race suite, lint, and hosted CI all pass.

umputun (approved):
lgtm.

both round-1 items check out, mutations included. Perf claim reproduces here, 5.1x wall time and 7.6x fewer allocs at cursor 5000.

one thing I checked and am fine with: the index is built over the whole file on every page motion, so motions near the top of a very large file pay ~11-24ms they didn't before, and everything past roughly line 500 gets much faster. Net win, no action needed.

PR #289, opened 2026-07-25
perf(ui): keep trackpad scrolling continuous

stevensuna:
## What changed

Diff-pane wheel packets update `YOffset` immediately. The existing 30 ms cursor-pin and `renderDiff()` debounce from #179 remains unchanged.

While wheel cursor work is deferred, `View()` reuses only stable pane chrome: the diff header, rendered tree or TOC pane, and status bar. The viewport and scrollbar remain live in every model view.

The Bubble Tea renderer now runs at its supported 120 FPS ceiling. Bubble Tea replaces buffered views between renderer ticks, so terminal-visible frames remain bounded by renderer cadence even when the model handles every packet.

## Why

The earlier 16 ms frame snapshot reduced repeated layout work but withheld intermediate viewport states before they reached Bubble Tea. Caching only stable chrome removes that model-level quantization while preserving the allocation improvement.

PTY tracing then exposed a second boundary: Bubble Tea's default 60 FPS renderer replaces buffered views between ticks. At an 8 ms packet cadence, master and the live-viewport branch both coalesced most intermediate offsets before terminal output. Raising the renderer ceiling to 120 FPS exposes more offsets without changing wheel scaling or the cursor-render debounce. Bursts faster than the 8.33 ms renderer interval can still coalesce.

## Benchmark

The benchmark runs the same raw `Update → View` workload on upstream/master and this branch using a 10,000-line wrapped diff with 1,000 annotations. Values are medians of five Apple M4 / Go 1.26.5 runs.

| Packets | upstream/master | This branch | Allocations |
|---:|---:|---:|---:|
| 1 | 0.52 ms, 648 allocs | 0.52 ms, 648 allocs | parity |
| 4 | 2.07 ms, 2,578 allocs | 1.94 ms, 1,822 allocs | 29.3% fewer |
| 16 | 8.41 ms, 10,292 allocs | 7.57 ms, 6,510 allocs | 36.7% fewer |

This measures model work, not terminal-visible frame count.

Raw PTY replay used a 541-line Markdown fixture with a TOC and 20 SGR wheel packets. A flush is counted from `CSI H` in the PTY byte stream, before any terminal emulator or multiplexer repaint behavior.

| Packet cadence | Zone | 60 FPS | 120 FPS |
|---:|---|---:|---:|
| 8 ms | content | 5–7 flushes | 10–11 flushes |
| 8 ms | TOC | 9 flushes | 16 flushes |
| 16 ms | content | 13 flushes | 17 flushes |
| 30 ms | content | 17–19 flushes | 20 flushes |

Flush count is an observed renderer-output metric, not an exact count of distinct model states.

## Verification

- `MISE_GO_VERSION=1.26.5 MISE_BUN_VERSION=1.3.13 mise exec go@1.26.5 -- go test -race ./...`
- `MISE_GO_VERSION=1.26.5 mise exec golangci-lint@2.11.4 -- golangci-lint run`
- focused live-viewport and cache-equivalence regression tests
- raw PTY SGR replay at 8, 16, and 30 ms cadences over content and TOC zones
- `git diff --check`

There are no CLI, config, or public API changes.

### DIFF
--- a/app/ui/loaders.go
+++ b/app/ui/loaders.go
@@ -526,8 +526,7 @@ func (m Model) handleFileLoaded(msg fileLoadedMsg) (tea.Model, tea.Cmd) {
 		return m, nil
 	}
 	m.file.requestedPath = ""
-	m.discardWheelFrame()
-	m.wheel.direction = 0
+	m.wheel.chrome = wheelChrome{}
 	m.wheel.renderPending = false
 	m.wheel.tickInFlight = false
 	m.wheel.gen++
--- a/app/ui/loaders.go
+++ b/app/ui/loaders.go
@@ -526,6 +526,11 @@ func (m Model) handleFileLoaded(msg fileLoadedMsg) (tea.Model, tea.Cmd) {
 		return m, nil
 	}
 	m.file.requestedPath = ""
+	m.discardWheelFrame()
+	m.wheel.direction = 0
+	m.wheel.renderPending = false
+	m.wheel.tickInFlight = false
+	m.wheel.gen++
 	if msg.err != nil {
 		m.layout.viewport.SetContent(fmt.Sprintf("error loading diff: %v", msg.err))
 		return m, nil
--- a/app/ui/model.go
+++ b/app/ui/model.go
@@ -975,6 +975,8 @@ func (m Model) Update(msg tea.Msg) (tea.Model, tea.Cmd) {
 		return m.handlePostFlushFinished(msg)
 	case wheelDebounceMsg:
 		return m.handleWheelDebounce(msg)
+	case wheelFrameMsg:
+		return m.handleWheelFrame(msg)
 	}
 
 	// forward other messages to textinput when annotating (e.g. cursor blink)
--- a/app/ui/model.go
+++ b/app/ui/model.go
@@ -975,8 +975,6 @@ func (m Model) Update(msg tea.Msg) (tea.Model, tea.Cmd) {
 		return m.handlePostFlushFinished(msg)
 	case wheelDebounceMsg:
 		return m.handleWheelDebounce(msg)
-	case wheelFrameMsg:
-		return m.handleWheelFrame(msg)
 	}
 
 	// forward other messages to textinput when annotating (e.g. cursor blink)
@@ -1339,12 +1337,7 @@ func (m *Model) toggleLineNumbers() {
 func (m Model) computeLineNumWidth() int {
 	maxNum := 0
 	for _, dl := range m.file.lines {
-		if dl.OldNum > maxNum {
-			maxNum = dl.OldNum
-		}
-		if dl.NewNum > maxNum {
-			maxNum = dl.NewNum
-		}
+		maxNum = max(maxNum, dl.OldNum, dl.NewNum)
 	}
 	if maxNum == 0 {
 		return 1
--- a/app/ui/model.go
+++ b/app/ui/model.go
@@ -591,7 +591,7 @@ type Model struct {
 	output      outputState       // transient hint state for the O in-session output flush
 	keys        keyState          // chord-pending state and transient hint for leader-chord keybindings
 	vim         vimState          // count accumulator, pending letter leader, and transient hint for vim-motion preset
-	wheel       wheelState        // diff-pane mouse wheel coalescing (debounced render via wheelDebounceMsg)
+	wheel       wheelState        // diff-pane wheel frame coalescing and deferred cursor/render state
 
 	ready        bool   // true after first WindowSizeMsg
 	filesLoaded  bool   // true after the first filesLoadedMsg is handled (keeps the loading view pinned until real data arrives)
--- a/app/ui/mouse.go
+++ b/app/ui/mouse.go
@@ -15,9 +15,9 @@ import (
 const wheelStep = overlay.WheelStep
 
 // wheelFrameDelay caps diff-pane wheel updates at the terminal's practical
-// render cadence. Raw trackpad packets accumulate until this tick fires, so
+// render cadence. Packets after the first accumulate until this tick fires, so
 // Bubble Tea can keep returning the previous frame instead of rebuilding the
-// full layout for every packet.
+// full layout for every raw trackpad packet.
 const wheelFrameDelay = 16 * time.Millisecond
 
 // wheelRenderDelay is the idle window after the last applied diff-pane wheel
@@ -26,27 +26,26 @@ const wheelRenderDelay = 30 * time.Millisecond
 
 // wheelState tracks the coalescing state for diff-pane wheel events.
 //
-// delta and frameInFlight collect raw packets behind one 16 ms tick. snapshot
-// holds the last complete frame so View stays cheap until the tick applies the
-// exact accumulated distance. frameGen invalidates ticks after an early flush.
+// The first packet in a burst moves the viewport immediately. targetOffset and
+// frameInFlight collect later same-direction packets behind one 16 ms tick.
+// snapshot holds the first packet's complete frame so View stays cheap while
+// those later packets accumulate. direction detects reversals even between
+// frames, and frameGen invalidates ticks after an early flush or a file switch.
 //
 // gen is bumped on every applied frame that shifts YOffset. renderPending stays
 // true while a cursor pin/render is owed (cleared by flushWheelPending).
 //
-// tickInFlight gates tea.Tick scheduling: only ONE tick is alive at a time,
-// regardless of how many wheel events arrive. The first wheel of a burst
-// schedules a tick; subsequent wheels in the same burst just bump gen. If the
-// tick fires with stale gen (burst still going), handleWheelDebounce
-// reschedules a new tick for the current gen. This keeps the debounce-msg
-// count proportional to burst duration (one per wheelRenderDelay), not to
-// wheel event count — drops thousands of redundant Update+View cycles on
-// trackpad/free-spin bursts and stops the wheel-up from queueing behind them.
+// tickInFlight gates render-debounce scheduling: only one debounce tick is alive
+// across the burst. Applied frames bump gen; a stale debounce tick reschedules
+// for the current gen. This keeps debounce traffic proportional to burst
+// duration rather than packet count.
 type wheelState struct {
 	gen           int
 	renderPending bool
 	tickInFlight  bool
 
-	delta         int
+	targetOffset  int
+	direction     int
 	frameGen      int
 	frameInFlight bool
 	snapshot      string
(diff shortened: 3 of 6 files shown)


umputun (commented):
the coalescing itself is sound, I traced the frame state machine and couldn't construct a state where `wheel.delta` or `wheel.snapshot` survives with no live tick. Perf claim reproduces here too: 6.72x wall time and 70.6% fewer allocs at 16 packets/frame, measured against master with your own benchmark. Nothing blocking, but a few things worth fixing first.

**1. single-notch latency**: `handleWheel:283` defers the apply to the frame tick even when a packet arrives alone in its window, so a discrete notched mouse waits 16ms plus up to 16.67ms of beat against bubbletea's own 60 FPS renderer ticker. That is 16-33ms before the pane moves, on a change whose point is responsiveness. The snapshot render on that path is discarded work too, it re-renders the pre-scroll model that is already buffered.

pls apply the first packet of a window immediately through the existing `scrollDiffViewportBy`/`renderPending` path and start accumulating from the second. `P=1` then matches master exactly and needs no snapshot at all, `P>1` still collapses to one applied frame.

**2. reversal at the edges**: `canScrollDiffViewportBy` compares the clamped target against the un-moved `YOffset`, so during accumulation `delta` grows past the real headroom. With `WheelStep=3` and 2 rows left below: four down packets accumulate `delta=12`, the reverse packet nets it to 9, the flush clamps to the bottom. Master ended 3 rows up. So the first up notch after crossing an edge produces no upward motion and jogs the view further down instead. Only the edge-crossing frame is affected, away from edges the clamp is the identity.

storing the clamped reachable target instead of the raw sum fixes it, and folds into 3 below. Worth a test that bursts into the bottom edge then reverses in the same frame, both existing reversal tests run mid-file so neither catches it.

**3. duplicated clamp**: `canScrollDiffViewportBy:321` re-derives `scrollDiffViewportBy`'s math byte for byte with nothing pinning the two equal, same shape as the `cursorVisualOffsets` one on #287. One `clampedDiffOffset(delta)` helper both call would fix it and hand the accumulator the clamped target 2 needs. It is also the only func in `mouse.go` without a godoc.

**4. pending frame survives a file switch**: `handleFileLoaded` (`loaders.go:588`) ends with `GotoTop()` but never touches the wheel state, so a live `wheelFrameMsg` fires ~16ms later and scrolls the freshly loaded file by the accumulated delta, then the debounce pins the cursor to that offset. The mouse and key flushes fire when a load starts, not when it completes, and `handleReviewFingerprintLoaded` starts loads with no keystroke at all. A `discardWheelFrame()` (zero `delta`, clear `snapshot` and `frameInFlight`, bump `frameGen`) right after the seq guard covers it.

**benchmark**: two problems. It never dispatches `wheelFrameMsg`, so it skips a full render the real message flow pays, which is exactly why `packets_per_frame_1` reads as parity when that case is the regression in 1. And `nav.diffCursor` is not reset per iteration, so `pinDiffCursorTo` fires only on the first one and the absolute numbers move ~90x with `b.N`. The A/B holds since both arms share the structure, but as a measuring stick to compare against later it will not.

**tests**: `TestModel_HandleMouse_WheelBurst_OppositeDirectionProcessesImmediately` lost the assertions covering the `tickInFlight` gate. Some of that churn was fair, wheel events no longer schedule debounce ticks, but the gate did not go away, it moved into `flushWheelFrame` and is now at 0 coverage. That is the one keeping ticker count proportional to burst duration rather than event count. Applying a second frame while `tickInFlight` is true and asserting the nil cmd covers it.

`WheelFramePreservesInlineAndFileAnnotations` cannot fail on the store or wrap assertions, nothing on the wheel path mutates either, and its one live assertion duplicates the existing one at `mouse_test.go:249`. `nav.diffCursor` and `cursorOnAnnotation` are what that annotated wrapped fixture actually exercises.

**docs**: `wheelState`'s comment still says wheels "just bump gen" six lines under the paragraph you did rewrite, and `handleWheel`'s still claims "the only per-event work is SetYOffset" and "each event is O(1)" when the path now runs a full `View()`. `flushWheelPending`'s enumerated call-site list says four, there are eight now, and its "no-op when no render is pending" is false since the frame teardown runs before that early return. `View()`'s godoc still says it renders the full TUI with no mention of the cached branch. `docs/ARCHITECTURE.md:116,131` also still has the three-field `wheelState` and the single-tick description, #179 updated that file in the same commit.

stevensuna:
The single-notch latency and edge reversal cases exposed two gaps in the original frame model: the first packet did not need coalescing at all, and a raw delta was not a safe representation once the viewport hit a boundary. I pushed b8767fc to address both.

The first packet now applies immediately with no snapshot or frame tick; the second packet starts frame coalescing. Later packets accumulate a clamped target through one shared offset helper, so reversing after reaching an edge moves away from it immediately. Accepted file loads invalidate both pending frame input and the old render-debounce state, preventing stale ticks from affecting the new file.

The benchmark now dispatches wheelFrameMsg and resets diffCursor and annotation cursor state on every iteration. Coverage now includes the in-flight debounce gate, clamped edge reversal, stale file-load ticks, and the actual wrapped-annotation cursor state. I also updated the wheel comments and architecture state description.

Local go test -race ./... and golangci-lint v2.11.4 pass.

umputun (commented):
one thing to fix first. As it stands this is slower than master in a common case, on a PR whose whole point is to make it faster.

`app/ui/mouse.go:293` - the snapshot is taken before the reachability check. Once the viewport is clamped at an edge, every further packet in that direction still enters the frame-start branch, runs a full `m.View()`, throws it away, and then bubbletea renders again after `Update` returns. Two full renders per packet until the debounce fires, where master returns at `scrollDiffViewportBy == false` and pays one. Flicking to the end of a file is the ordinary way a scroll ends, so this is the common path, not a corner.
same wasted render you flagged in point 1, moved off the first packet onto the at-edge path - af6bf67 called `canScrollDiffViewportBy` first, and the rewrite that introduced `clampedDiffOffset` inverted the order.
Fix: compute the clamped target first, return when it equals `YOffset`, then take the snapshot. The `m.wheel.snapshot = ""` rollback becomes dead.

**open question**: the eager snapshot on the second packet pays off only when a third arrives inside the same 16 ms. At roughly one packet per window, a notched wheel spun hard or the inertial tail, it costs snapshot plus frame-apply where master did one render. Worth holding the snapshot back until a second packet is actually there, or is the flick case the only one you care about?

the rest of last round's points check out.

stevensuna:
The open snapshot question exposed a broader UX problem: even when batching improved throughput, retaining the prior complete frame made continuous trackpad input visibly step at the 16 ms boundary. I pushed 611cb5c to remove that trade-off.

Every wheel packet now updates the live viewport immediately. Only stable chrome is cached during the existing cursor-render debounce, and regression coverage compares every cached frame byte-for-byte with the uncached render. A PTY E2E run injected raw SGR wheel packets into the built binary and observed top rows 1 → 4 → 7 → 4 for down, down, up, with a clean exit.

On the same 10,000-line wrapped fixture, the 16-packet path is 7.57 ms versus 8.41 ms on upstream/master and allocations fall from 10,292 to 6,510. Full race tests and lint pass with the pinned mise toolchains.

umputun:
I'm going to close this one, and the reason is the problem statement rather than your execution.

there is no reported symptom behind it. #179 removed the expensive per-packet work and nobody has come back with a follow-up complaint. I measured master against this branch on a 10k-line wrapped file with 1000 annotations: 8.75 ms vs 7.72 ms for a 16-packet burst, 545 vs 544 µs for a single packet. Both sit inside one 16.7 ms frame, so the difference is not something a reviewer feels. The chrome the cache owns is about 17 µs of a ~483 µs packet, so roughly 88% of the per-packet cost is the diff-pane render chain that no version of this touched.

the other half is that three rounds produced three designs: frame-tick accumulation, then a whole-frame snapshot, then the chrome cache. That is partly on me. My question about the eager snapshot is what sent you into the round-3 rewrite, and you answered it in good faith. But a change that keeps changing shape under review is a sign the target was not well defined in the first place, and I should have worked that out before the first round instead of the third.

where it ended up still has costs I would be taking on permanently:

`app/ui/view.go:147` - `prepareWheelChrome` re-derives `diffPaneW` and hand-copies the whole left-pane build from `renderTwoPaneLayout`. Two render paths that have to stay byte-identical forever, and `.claude/rules/gotchas.md` already pins that width derivation to exactly two sites.

`app/ui/view.go:143` with `app/ui/loaders.go:420,616` - the chrome cache is only cleared by `flushWheelPending` and `handleFileLoaded`. `handleFilesLoaded`'s error and empty-entry returns, and `handleReviewFingerprintLoaded` when the selection does not change, leave a stale tree and reviewed count on screen for the rest of the burst. Reachable with `u`, or by marking an unloaded file reviewed.

`app/ui/mouse.go:36` and `app/ui/view.go:141` - the cache godoc says these elements "cannot change" during a burst. `tree.Rebuild` and `tree.SetReviewed` both change them.

`app/ui/mouse_test.go:1265` - the cached-vs-uncached comparison cannot fail. Strip every `wheel.chrome` read out of `View()` and the whole suite still passes, because the fixture uses a plain resolver, a one-entry tree, and identical styles for both tree pane keys, so the scrollbar, padding and focus-style calls are no-ops on both sides.

smaller ones, for completeness: `View`'s godoc lost the cached-branch sentence you added in b8767fc; `handleWheelDebounce:306` still names three flush paths out of eight; `mouse.go:170` flushes on the horizontal-wheel arm, which reads nothing; `renderDiffHeader` and `renderStatusBar` are the only two methods in view.go without a godoc; `updateWheelAndFlush` no longer runs the returned cmd, so the tick-scheduling contract is unpinned; and the two `max()` rewrites in `model.go` and `view.go` are unrelated to the wheel path.

none of this is wasted if a real case turns up. If someone reports scrolling that actually stalls on a large diff, reopen it and point me at the repro, and the right fix then is probably the ~88% rather than the chrome. #287 was a clear win and I merged it without much argument, so please read this as a call on this particular change and not a signal about your work here.

stevensuna:
The PTY trace found a renderer boundary beyond the model-level caching change. Bubble Tea's standard renderer replaces buffered views between its 60 FPS ticks, so the previous claim that every wheel packet produced a terminal-visible frame was incorrect.

`e87195e` adds `tea.WithFPS(120)`, the renderer's supported ceiling. With 20 raw SGR wheel packets, observed content flushes increased from 5–7 to 10–11 at 8 ms cadence, from 13 to 17 at 16 ms, and reached 20 at 30 ms. TOC traffic at 8 ms increased from 9 to 16 flushes. Faster-than-8.33 ms bursts can still coalesce.

I updated the PR body to separate the verified allocation benefit from terminal continuity and to describe flush count as a renderer-output metric rather than an exact model-state count. Full race tests, lint, build, and repeated PTY replay pass.

PR #290, opened 2026-07-26
fix(launcher): hand review to a live tmux session when called headless

kylesnowschwartz:
> **Draft — still in local testing.** The logic is covered by an isolated automated test suite (details below), but I have not yet run it through a real headless agent handoff on my own machine. I will take it out of draft once that manual QA passes. Review welcome in the meantime.

## Problem

A background or headless agent session (no controlling terminal, and none of the client environment variables like `$TMUX`, `$ZELLIJ`, `$KITTY_LISTEN_ON`) cannot launch a revdiff review. Every overlay backend in `launch-revdiff.sh` gates on a client-inherited environment variable, so a session with no terminal ancestry always falls through to:

```
error: no overlay terminal available (requires agterm, tmux, zellij, ...)
```

This happens even when a tmux server is running on the default socket with an attached, focused client that could have hosted the review. Background/headless agent sessions are exactly the case where the agent cannot render a TUI itself and most needs to hand the review to the user's live terminal.

**Repro:** from any shell with `$TMUX` unset but a running, attached tmux server, run the launcher with a valid ref. Before this change: exit 1, "no overlay terminal available".

## Fix

When `$TMUX` is unset but `tmux` is on `PATH`, the launcher now probes the default server. If a client is attached, it opens the review in a **server-owned tmux window** (which survives a client disconnect, unlike a `display-popup`) addressed with an explicit `-t` at the most-recently-active attached client's session — so nothing depends on tmux's ambiguous client-less "current session" resolution.

- `REVDIFF_TMUX_TARGET=<session>` forces this path at an explicit session and skips the probe. It also works from inside tmux to target another session.
- The path falls through untouched when there is no tmux, no server, or no attached client — the later backends and the original final error still apply.
- The window backend (`agentdeck-window.sh`) now threads the explicit target through the prior-window capture, `new-window`, the watch loop (`list-windows -a`), and the prior-window restore. With no target it is byte-for-byte the previous behavior (agent-deck auto-detection and a bare `REVDIFF_TMUX_WINDOW=1` are unchanged).

Both are launcher environment variables, not revdiff flags.

## Scope

- `launch-revdiff.sh` + `agentdeck-window.sh` (Claude plugin), with the Codex plugin copies kept in sync.
- Docs: `README.md`, `site/docs.html`, `SKILL.md` (both plugins), and the `config.md` / `usage.md` references document the fallback and the new `REVDIFF_TMUX_TARGET` variable.

## Testing

- `shellcheck -x` clean on all four scripts; parses under bash 3.2 (macOS system bash) and bash 5.
- An isolated automated suite (a `tmux -L rdqa` sandbox server, a stub `revdiff` that reports which session hosted its window, run 3×) covers: no server → falls through; server but no attached client → falls through; attached client → auto-probe lands the review in that session; `REVDIFF_TMUX_TARGET` overrides the probe; the window closes on exit; and focus/restore returns the prior active window. 7/7 green, stable across runs.

Still to do before undrafting: a real headless-agent → live-terminal handoff on my machine.

### DIFF
--- a/.claude-plugin/skills/revdiff/SKILL.md
+++ b/.claude-plugin/skills/revdiff/SKILL.md
@@ -146,6 +146,8 @@ The resolver and launcher MUST run in the same bash invocation — the resolver 
 
 **Disconnect-resilient tmux window mode**: when running under tmux, prefix the launcher with `REVDIFF_TMUX_WINDOW=1` to open revdiff in a persistent, server-owned tmux window instead of a client-owned `display-popup`. The review then survives a dropped SSH or tmux client — reattach and it is still there. This is a launcher environment variable, not a revdiff flag.
 
+**Headless review handoff**: from a background session with no terminal of its own (no `$TMUX`), the launcher probes the default tmux server and, if a client is attached, opens the review as a server-owned window in that client's session instead of failing — no action needed. When the probe cannot pick the right session (multiple sessions, or none attached), prefix the launcher with `REVDIFF_TMUX_TARGET=<session>` to target one explicitly; it forces window mode and skips the probe. This is a launcher environment variable, not a revdiff flag.
+
 The script:
 - Detects available terminal (tmux → Zellij → herdr → kitty → wezterm/Kaku → cmux → ghostty → iTerm2 → Emacs vterm)
 - Launches revdiff in an overlay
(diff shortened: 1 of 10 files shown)


kylesnowschwartz:
Closing this — on reflection it treats the symptom, not the root cause.

The real issue is upstream of the launcher: the review request reached an agent session with no terminal of its own, and the skill tried to launch an overlay anyway. This PR's auto-probe makes that blind launch *succeed* by picking whatever tmux client happens to be attached — which can drop the review into an unrelated task's session. That trades a loud, correct failure for a quiet, surprising one.

The right fix belongs in the calling skill (detect a terminal-less context and hand the review back to the user's own terminal / defer) rather than in the launcher guessing a destination. Keeping the branch around; the deterministic `REVDIFF_TMUX_TARGET` hand-off may return later, paired with skill guidance that sets it deliberately.

PR #292, opened 2026-07-29
Relay revdiff's stderr when an overlay review fails

kylesnowschwartz:
When revdiff fails fast inside a terminal overlay — a bad ref, or a flag an older installed binary does not recognise — the overlay closes before anyone can read the error. The caller gets a bare nonzero exit with no explanation, and an AI agent driving the launcher has nothing to report but "it failed", so it guesses which argument was at fault.

This was built as part of #263 and does not depend on that PR's tmux-backend question, so it lands here on its own.

## Why this shape

Not a cherry-pick of #263's version. That one hung the capture off each backend's sentinel file, which silently excludes the backends that have no sentinel. Capturing into one file named up front needs no sentinel and covers everything, because every backend already funnels through the same two points: they all run the assembled command string, and they all exit through the same function. One redirect and one replay block, no per-backend edits.

**Relaying only on failure is the part worth arguing about.** Exit `0` is a clean quit and `10` means annotations were captured. Both are successes, and revdiff writes ordinary warnings to stderr — a deprecated keymap alias, for instance. A relay that fired on those would print stderr on nearly every successful review, which is a worse problem than the one being fixed. Hence the two negative tests.

## Confidence

Running the stderr expectations against the pre-change launcher fails on every sentinel backend: the error text was lost entirely, which is the failure mode the two-point design fixes.

Also checked by hand in a real tmux session, popup and window mode: a bogus flag now surfaces revdiff's own message instead of a bare `1`, and neither a clean quit nor an annotated review puts anything on stderr.

One backend is worth flagging for review: the sourced agent-deck window path is the only one absent from the test harness, so it is verified by hand rather than by test. It needed no code of its own.

## One deliberate behaviour change

Warnings that previously painted into the overlay are now discarded on a successful review. No backend passed revdiff's stderr back to the caller, so nothing an agent could read is lost — but the text did land on the popup's own screen, interleaved with the TUI and corrupting the render. Capturing it is the better trade, and it does mean a warning like a theme failing to persist now leaves no trace at all. Flagging it rather than hiding it in the diff.

## Out of scope

The planning launcher has the same blind spot but no shared helpers to hook into — fixing it properly means extracting those first, so it is left for a follow-up. No version bumps.


umputun (commented):
two things before this goes in.

`app/plugin_exit_code_test.go:165` - the new sweep is a second full launcher x backend pass, and it puts the `app` package over the `-timeout=60s` in `Makefile:29`. `make race` panics with `test timed out after 1m0s` inside `TestShellLaunchersRelayOverlayStderr`, 60.8s here against a 53.9s baseline with the test skipped. CI doesn't set that timeout so it stays green, only the local target breaks. Fix: fold the stderr expectations into `TestShellLaunchersPreserveAnnotationExitCode`, which already walks the same matrix with the same `fakeLauncherEnv`. `wantStderr` has to be gated per launcher since `launch-plan-review.sh` has no relay. Anything that keeps `./app` under 60s works.

`.claude-plugin/skills/revdiff/references/install.md:67` - the custom-launcher contract still lists only the three obligations and says nothing about the relay, while both SKILL.md edits now tell the agent the failure text will be there. An override built to the documented contract relays nothing, and one that adds a branch with its own EXIT trap leaks a `revdiff-err-*` per run. The same sentence was updated in 29d16c3, the PR that introduced the exit-10 contract. Fix: add the relay clause to the obligations sentence. Claude copy only, the codex `references/install.md` has no override section.

on the 0/10 exclusion you flagged for argument: it holds. I checked every backend and none inherited the launcher's stderr before this change, so the warnings now discarded on success were not reaching anyone to begin with.

kylesnowschwartz:
Both fixed in 5568fad.

**Test timeout.** Folded into `TestShellLaunchersPreserveAnnotationExitCode` and deleted the separate sweep. Launchers gain a `relaysStderr` flag — true for the two `launch-revdiff.sh` copies, false for `launch-plan-review.sh`, which also keeps `FAKE_STDERR` unset there so the plan launcher's cases are unchanged. Cases gain `wantStderr`, set on `failure` only.

That puts the launcher subprocess count back at the pre-PR matrix: 165 invocations before the fold (3x11x3 + 2x11x3), 99 after. I can't demonstrate the 60s target directly — this machine runs about 1.3x slower than yours, and `./app` is 69.4s on `master` with no changes at all — so what I measured is parity: 69.2s on the branch against that same 69.4s baseline, i.e. the fold costs nothing over master.

Guard strength is unchanged. Running the folded test against `81ddb9c^` still fails: the eight sentinel backends fail the `failure` case with empty stderr, so the relay assertion is still doing its job from inside the merged test.

**Custom-launcher contract.** Added the relay clause to the obligations sentence plus a line about naming the capture file in any added `EXIT` trap. Claude copy only — grepping for that sentence confirms `install.md` is the only file carrying it.

**On stderr inheritance:** you're right and I've reworded the PR body. I checked what my earlier "direct backends leaked it through" note actually measured, and it was the fixture, not the real path — `fake-overlay-backend.sh` runs the command as a plain child of the launcher, so tmux/ghostty/agterm inherit stderr in the harness where a real `display-popup -E` puts it on the popup's pty. Nothing was reaching the caller pre-change, as you said. The one thing that did change is on-screen: those warnings used to paint into the popup and interleave with the TUI, and now they don't. That's the render corruption the capture fixes, not a channel anyone was reading.

PR #297, opened 2026-07-30
feat: add opt-in pprof server

### LINKED ISSUE #296: Add opt-in pprof server for performance diagnostics
Add an opt-in `--pprof` mode so CPU and heap profiles can be captured from a running revdiff session.

Proposed behavior:
- Profiling stays disabled by default.
- Bare `--pprof` listens on `:6060`.
- Custom addresses use `--pprof=ADDR`; this preserves the following positional argument as the diff ref.
- Bind failures are logged and remain non-fatal.

quetz:
## Summary

- add an opt-in `--pprof` debug server for CPU and heap profiling
- use `:6060` for bare `--pprof` and require `--pprof=ADDR` for custom addresses
- keep bind failures non-fatal and document the new flag

## Testing

- `go test ./... -count=1`
- `golangci-lint run`

Closes #296

### DIFF
--- a/.claude-plugin/skills/revdiff/references/config.md
+++ b/.claude-plugin/skills/revdiff/references/config.md
@@ -59,6 +59,7 @@ Then uncomment and edit the values you want to change.
 | `--dump-keys` | | Print effective keybindings to stdout and exit | |
 | `--config` | `REVDIFF_CONFIG` | Path to config file | `~/.config/revdiff/config` |
 | `--dump-config` | | Print default config to stdout and exit | |
+| `--pprof` | | Start a pprof http server for CPU/heap profiling at ADDR; bare `--pprof` binds `:6060`, use `--pprof=ADDR` for a custom address | |
 
 Config-backed options use long flag names without leading `--`; for annotation exit status use `exit-code-on-annotations = true`.
 
--- a/README.md
+++ b/README.md
@@ -400,6 +400,7 @@ Positional arguments support several forms:
 | `--dump-keys` | Print effective keybindings to stdout and exit | |
 | `--dump-config` | Print default config to stdout and exit | |
 | `-V`, `--version` | Show version info | |
+| `--pprof` | Start a pprof http server for CPU/heap profiling at ADDR; bare `--pprof` binds `:6060`, use `--pprof=ADDR` for a custom address | |
 
 ### Config File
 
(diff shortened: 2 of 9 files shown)


umputun:
going to decline this one, with the reasoning rather than a silent close.

nothing here explains what the actual problem is or why this code is needed. Issue #296 describes the mechanism ("bare `--pprof` listens on `:6060`"), not what is slow or broken, and the commit on this branch is authored 06:22Z while the issue was filed 07:58Z. The code came first and the issue was written around it. CONTRIBUTING.md asks first-time contributors to open an issue and wait for approval before writing code, and separately asks every PR to state what the problem is. Same call I made on #236.

beyond process, three things about the change itself.

bare `--pprof` binds `:6060`, which is every interface, not loopback. The tool runs on a laptop and holds the diff of whatever repo you point it at, so that default publishes `/debug/pprof/heap` and `/debug/pprof/cmdline` (argv, i.e. `--output` and `--description-file` paths) to whatever network the machine happens to be on. The only defensible default here is `127.0.0.1:6060`.

it also puts the first HTTP listener into the binary, and that is not free. There is no `net/http` in revdiff today at all, so pulling in `net/http/pprof` grows the release build from 8.48 MiB to 12.07 MiB, up 42%, for every user who will never touch the flag. A terminal tool taking a 3.6 MiB hit and a network socket for a maintainer-only diagnostic is not a trade I want to make.

the `optional-value` string flag misparses without an error:

```
--pprof :7070                  -> Pprof=":6060"  base=":7070"
--pprof 127.0.0.1:7070 HEAD~3  -> Pprof=":6060"  base="127.0.0.1:7070"  against="HEAD~3"
```

only `--pprof=:7070` sets the address. `app/config_test.go:881` pins the benign direction, but the address silently becoming the diff ref is the real cost of making this a string flag instead of a bool.

and I am not convinced it would tell you much. Bubbletea blocks on input, so a CPU profile of a session that spends most of its life idle in a read is mostly noise. The two actual perf problems here, #179 and #287, were found by counting View cycles and measuring row math, not with pprof.

if you want profiling, the shape that would fit is `runtime/pprof` writing a CPU or heap file on exit. No listener, no port, no bind address, and no size hit worth mentioning. Open an issue for it first that explains why it is needed rather than how it works, and we can weigh whether it is worth carrying.

quetz:
Reason - `revdiff` is unusable on large files because it redraws whole diff on every keystroke.  To find that out I needed `pprof` support inside it. It could've been one-off patch, but I thought it will be useful in future debugging sessions.

Writing a CPU or heap profile on exit is worse option because I needed to profile how app behaves while typing out annotation, not end-to-end behaviour.

On binding address and parsing ambiguity - fair points, maybe `--pprof` argument should be required.

Issue #281, opened 2026-07-19
Allow annotating description


chris-peterson:
## What is the problem?

The info popup (`i`) reveals [`--description`](https://github.com/umputun/revdiff#review-description) (commit message(s) and/or other context behind a change), as _read-only_.

Being able to make adjustments to _intent_ with the same review ergonomics as the _outputs_ would be powerful.

## Why it matters

Correcting an agent's context is a high-leverage edit. A wrong assumption in context keeps producing the same class of mistake in the code, so one correction to the intent is far-reaching; often spurring the agent to make multiple code corrections.  Even better, the session is then better tuned for subsequent turns.

Fact-checking in-session context before it calcifies is important.  Agents often read back git blame or raw commit messages to fill in context gaps; a committed inaccuracy is a future liability.

## Implementation suggestions

Use a synthetic "file" for the description, e.g. `(description)` that would be anchored at the top of stdout

umputun:
I don't think this fits revdiff, but I don't want to make it a hard no either.

the mental model here is deliberately simple: the artifacts you review are diffs, and annotations belong to the diff pane. The description in the `i` popup is context I (or an agent) supply as input, not an artifact under review, so annotating it sits outside that model, and it'd be a new interaction with no precedent, today every annotation lives on a diff line.

most of the intent-level feedback you're after is already reachable. Nothing stops you from putting a higher-level annotation on any diffed file, line or file-level, about things beyond the file itself, including a wrong assumption in the intent. If you specifically want to annotate standalone prose on its own, `--stdin` already gives line-by-line annotations of arbitrary text with the same structured output.

there are technical hurdles too. The description is one read-only, pre-highlighted string with no line numbers or change types, while every annotation keys on file:line:type. Wiring it in means either making that popup editable, or promoting the description into the tree as a synthetic file that then has to be special-cased out of fingerprinting, stats, filters, compact/collapsed/blame, and the `## file:line` output that agents parse as real paths.

I also doubt the reach, not sure the `i` box gets opened by many people, so this feels niche.

so I'm not sold on the need or the fit. But I'm open to being shown, if someone puts up a PR that demonstrates it well and keeps it simple enough for a niche feature, I'll consider it.

umputun:
closing this one. #301 took up the offer and gave the idea a fair shot, reasoning is there.

[...]

PR #301, opened 2026-08-02
feat: edit the review description from the info popup

chris-peterson:
## Context

Along with a diff, prose context can be provided via `--description`
(this can be a prior commit message, an agent's intention, etc)

This context can currently be viewed via an info popup (`i`)

This PR enables editing this context.  Correcting _intent_ is a high-leverage edit: a wrong
premise keeps generating the same class of mistake, so one fix to the premise can often
flow to multiple change sites, avoid repetitive corrections, or even reduce session turn count

## Demo

A `/ship-it` run opens an **already-committed** change for review, passing the
commit message as the review context

The reviewer annotates **no code at all**. They open the context (`i`), press
`e`, and correct the one assumption. On exit the correction comes back as a
`## (description)` block, and the skill acts on it: two code defects that both
followed from the premise (no idempotency key on a replayed charge; a
`time.Sleep` backoff that ignores context cancellation), plus the commit message
that asserted it. A second pass returns no annotations, and that clean signal is
what releases the change

One context correction driving two code fixes and a commit-message rewrite

Note the popup mid-clip: the correction renders under a `description (edited)`
header while `commits (1)` still shows the original, uncorrected commit beneath
it

> The assistant turns in the clip are scripted. The two revdiff passes, the git
> state, the file edits, and the `--amend` are real; the narration around them
> is staged

## Review guide

- [exported const](https://github.com/chris-peterson/revdiff/pull/4/changes#diff-87297acfa6e391df0db42099534ce377685a2b0860e8bffa3b5ebcbf2cead7df-R16)
- [`annotations_load.go` R126](https://github.com/chris-peterson/revdiff/pull/4/changes#diff-f4a0635bfe3d23a72ad96a3bd6788c4e26a02a2c519b74c8bc1c9a8bb4c0a321-R126)
  — the `--annotations` preloader resolves every record against the diff and
  drops what it can't find. Without this exemption an `-o` snapshot silently
  loses its description edit on reload, which is exactly the leak you predicted
  on the issue
- [`info.go` R152-154](https://github.com/chris-peterson/revdiff/pull/4/changes#diff-837eb274bbff89054162a594e10dc5892ddfd806b06ad84ca16daaa6d1770a9b-R152)
  — the edit **replaces** the supplied prose under a `description (edited)`
  header instead of rendering as a second section, so a corrected assumption is
  never displayed next to the wrong one it supersedes

The rest, in descending order:

- [`reviewinfo.go` R208](https://github.com/chris-peterson/revdiff/pull/4/changes#diff-421ec9fee1006bc6094853874515487c3cc88aafcdf2773285d54b864c9769be-R208)
  and [R226](https://github.com/chris-peterson/revdiff/pull/4/changes#diff-421ec9fee1006bc6094853874515487c3cc88aafcdf2773285d54b864c9769be-R226)
  — open `$EDITOR` seeded from the existing edit, else the description prose;
  store or (on an empty buffer) delete on return. Error handling mirrors
  `handleEditorFinished`: an unchanged read-back on error counts as no edit.
- [`model.go` R1258](https://github.com/chris-peterson/revdiff/pull/4/changes#diff-f35be84b66d47d2dc83baabcf787c38dc460cc4a3e712310d57af4d761b09e07-R1258)
  — the key is gated in the Model, not the overlay, so `app/ui/overlay` stays
  free of the annotation store
- [`reviewinfo.go` R164](https://github.com/chris-peterson/revdiff/pull/4/changes#diff-421ec9fee1006bc6094853874515487c3cc88aafcdf2773285d54b864c9769be-R164)
  — the popup hint is composed from the keymap rather than naming a literal
  `e`, since `open_file_in_editor` is rebindable. Chords are excluded because
  chord dispatch goes through `dispatchAction` and never reaches the popup's
  gate in `handleModalKey`, so a chord-only binding genuinely cannot annotate
  the description
- docs across `README.md`, `site/docs.html`, and both plugin `usage.md` refs,
  including a note telling agents that `## (description)` is not a file path

## Approach & trade-offs

**Reusing `open_file_in_editor` rather than adding an action.** It already means
"edit this in `$EDITOR`" and it's unbound while the info popup is open, so the
binding is free and no new keymap surface appears. Cost: a user who rebinds it
moves this too, which is why the hint follows the keymap

**Seeding the first edit with the description text** rather than a blank buffer.
It makes the edit a revision of the context, which is what the replace-on-display
behavior above then reflects. Every other annotation starts blank, so this is the
one place the feature departs from the existing pattern

### DIFF
--- a/.claude-plugin/skills/revdiff/references/usage.md
+++ b/.claude-plugin/skills/revdiff/references/usage.md
@@ -85,6 +85,8 @@ revdiff HEAD~3 --description-file=/tmp/review-notes.md
 
 `--description` and `--description-file` are mutually exclusive. Both are optional — omit when there's no useful context to add.
 
+The reviewer can edit the description: pressing `e` while the info popup is open opens `$EDITOR` seeded with the description text, emitted as a `## (description) (file-level)` block (see Output Format). The edit replaces the prose shown in the popup, so the corrected assumption supersedes the original rather than sitting beside it. This is how a reviewer corrects a wrong assumption in the context you supplied.
+
 ## Two-File Diff
 
 Use `--compare-old=<path>` together with `--compare-new=<path>` to diff two arbitrary files on disk using `git diff --no-index`. No VCS repo needed — works anywhere `git` is installed.
@@ -182,6 +184,7 @@ Press `Space` to mark the focused file reviewed. Press `F` to toggle the sidebar
 | `F` | Toggle filter: all files / unreviewed only |
 | `?` | Toggle help overlay showing all keybindings |
 | `i` | Toggle info popup — review scope (mode, VCS, ref, filters, file/status counts, aggregate `+/-` stats) plus the commit log for the current ref range when applicable |
+| `e` (in info popup) | Edit the `--description` in `$EDITOR`; replaces the displayed prose, emitted as `## (description) (file-level)` |
 | `R` | Reload diff from VCS (warns if annotations exist) |
 | `q` | Quit, output annotations to stdout |
 | `Q` | Discard all annotations and quit (confirms if annotations exist) |
@@ -303,6 +306,8 @@ don't remove this validation
 ```
 
 Each annotation block: `## filename:line[-end] (type)` where type is `(+)` added, `(-)` removed, or `(file-level)`. The `-end` suffix is included when the annotation covers a line range.
+
+A `## (description) (file-level)` block is not a file path: it is the reviewer's annotation on the `--description` you supplied, added with `e` in the info popup. Treat it as a correction to the intent/context of the change, not a comment about a file named `(description)`.
 
 When annotation text contains the keyword "hunk" (case-insensitive, whole word), the output header automatically expands to include the full hunk line range (e.g., `handler.go:43-67 (+)` instead of `handler.go:43 (+)`). This gives AI consumers the range context without any extra steps.
 
--- a/README.md
+++ b/README.md
@@ -661,7 +661,9 @@ See ticket SEC-441 for context."
 revdiff HEAD~3 --description-file=.review-description.md
 ```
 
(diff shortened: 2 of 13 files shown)


umputun (commented):
going to pass on this one.

on the technical half of #281 the hurdles I listed are actually gone. No synthetic tree entry, so fingerprinting, stats, filters and compact/collapsed/blame are untouched, and the `--annotations` orphan-drop that would have eaten the note on reload is closed at `app/annotations_load.go:123`.

what isn't answered is the objection I led with. The description is context supplied as input, not an artifact under review. A synthetic `(description)` key plus a reused `open_file_in_editor` binding doesn't make it one, it produces an annotation that doesn't sit on a diff line, and every annotation today does. The PR works around that point rather than arguing it.

the seeding choice also breaks the loop this tool exists to feed:

`app/ui/reviewinfo.go:568` - the buffer is seeded with the description, so `:wq` with no edit stores a note identical to my own prose. Exit code is 10 and the calling agent gets its own text back as reviewer feedback. Only the error path checks `content == seed`.

`app/ui/model.go:1304` - a `(description)` note preloaded through `--annotations` into a run with no `--description` is hidden by the `descriptionHighlighted != ""` gate, un-jumpable in `@`, and can't be deleted. The store stays non-empty so exit stays 10.

`.claude-plugin/skills/revdiff/SKILL.md:177` and `plugins/codex/skills/revdiff/SKILL.md:206` still specify `## filename:line (type)`. Only the two `references/usage.md` got the carve-out, and `plugins/pi/` got nothing.

the two paths I pointed at on the issue still cover the case: a file-level annotation on any diffed file saying the premise is wrong, or `--stdin` on the prose itself. Both work today, and neither adds permanent surface for something I still read as niche.

Issue #307, opened 2026-08-05
iTerm2 overlay pane is never named, blanking the tab and window title


chris-peterson:
## What is the problem?

While a review is open in iTerm2, the tab hosting it stops identifying its session, and the OS window title empties out with it. Both come back when the overlay closes.

Both surfaces read the **active** session's name, and the overlay pane is created without one of its own. `split … with same profile` copies the parent session's profile — including the session-scoped `name` override iTerm2 stores a session name as — but not the parent's session variables. So where the parent's name interpolates variables (`\(user.…)`, the only way to keep a label current on a profile that disables title-setting OSCs), the copy has nothing to interpolate.

Measured on iTerm2 3.6.11, with the parent's name set to `<b>\(user.probe_var)</b>` and `probe_var` published in the parent:

```
parent   name=[<b>PARENT</b>]   var=[PARENT]
overlay  name=[<b></b>]         profileName=[<b>\(user.probe_var)</b>]   var=[missing value]

overlay open, not yet focused   tabTitle=[<b>PARENT</b>]   windowTitle=[PARENT]
overlay focused                 tabTitle=[<b></b>]         windowTitle=[]
```

Where the parent's name is a static string, the pane inherits that string instead, and the tab spends the review claiming to show something it isn't.

iTerm2 is the only backend whose overlay goes unlabelled. `OVERLAY_TITLE` (`rd: <dir> [ref]`, `launch-revdiff.sh:116`) is applied for tmux (`-T`), zellij (`--name`), herdr (`--label`), kitty (`--title`), and the emacs frame name. The iTerm2 branch never applies it to the session it creates, and nothing in `revdiff --help` sets it either.

## How does this solve it?

Name the session the split returns, using the title the launcher already computed:

```applescript
set newSession to split horizontally with same profile command cmd
set name of newSession to overlayTitle
```

That addresses the cause — a pane with no name of its own — rather than the symptom, and puts iTerm2 on the footing the other backends already have: the tab reads `rd: revdiff [HEAD~1]` for the duration of the review and reverts on its own when the pane closes, since the name belongs to the session.

There is no tab-level alternative to fall back on: `tab` exposes a `title` property in iTerm2's scripting dictionary, but setting it fails with `AppleEvent handler failed`.

### What would need to change

| File | Change |
|:---|:---|
[...]

PR #309, opened 2026-08-05
Name the iTerm2 overlay pane so the tab keeps its label

chris-peterson:
## Context

The iTerm2 overlay opens as a split pane, and that pane is created with no name of its own. A tab's label comes from its *active* session, so once the overlay takes focus the tab falls back to the profile's title for the duration of the review and stops saying which session it belongs to. #307 carries the measurements.

tmux (`-T`), zellij (`--name`), herdr (`--label`), kitty (`--title`) and the emacs frame name already label their overlay from `OVERLAY_TITLE`. iTerm2 is not the only backend that doesn't — wezterm, cmux, ghostty and agterm don't either — it is the one whose pane visibly relabels the tab. This passes the same title into the iTerm2 branch and sets it on the session the split returns. Nothing has to undo it afterwards: the name belongs to the session, so it goes when the pane does.

Reach is narrow: on a stock profile the parent and the overlay render the same title and the tab doesn't change. This shows up for people who name their sessions.

## Review guide

- **The fix** — `set name of newSession to overlayTitle` after the split, with `$OVERLAY_TITLE` appended to the `osascript` argv, in `.claude-plugin/skills/revdiff/scripts/launch-revdiff.sh`
- **The same edit in the codex copy**, which carries the `# source: … (keep in sync)` header
- **And in `plugins/revdiff-planning/scripts/launch-plan-review.sh`** — the planning hook splits the same way and computes its own `OVERLAY_TITLE` (`plan: <file>`), so plan reviews had the same unlabelled tab
- **Why the title is appended rather than inserted** — the `osascript` arm of `app/testdata/plugin-exit-code/fake-overlay-backend.sh` dispatches on argument count plus `-x "$3"`, so an earlier position would shift the launch script out from under the stub. Appending leaves the stub's dispatch intact.
- **The test** — `TestIterm2OverlayNamesSession` asserts the split `osascript` argv ends with the overlay title for all three launchers. The other backends pass `OVERLAY_TITLE` as a flag that `TestShellLaunchersPreserveAnnotationExitCode` already runs; iTerm2 hands it to AppleScript through argv, which nothing read, so a dropped argument was invisible. Capture is a new `FAKE_OSASCRIPT_ARGS_FILE` in the fake backend, mirroring `FAKE_AGTERM_ARGS_FILE`.

## Testing

Ran the launcher for real on iTerm2 3.6.11 / macOS 26.5.2, against a profile whose session name interpolates a session variable:

| | overlay pane name | tab label |
|:---|:---|:---|
| `master` | the profile's title, which renders empty here — the probe in #307 shows the pane carrying `<b>\(user.probe_var)</b>` with the variable missing. On a stock profile it is the default title text. | the same, once the overlay takes focus |
| this branch | `rd: revdiff [HEAD~1]` | `rd: revdiff [HEAD~1]` |

The tab label is what the fix restores. My OS window title emptied along with it, but @umputun's never moved, so that part follows from window title components rather than from the session name.

`go test ./...` passes and the CI shellcheck gate is clean on all three launchers.

### DIFF
--- a/.claude-plugin/skills/revdiff/scripts/launch-revdiff.sh
+++ b/.claude-plugin/skills/revdiff/scripts/launch-revdiff.sh
@@ -540,9 +540,10 @@ on run argv
                                 set newSession to split horizontally with same profile command cmd
                             end if
                         end tell
-                        -- the tab label and window title come from the active
-                        -- session's name, and a split inherits the parent's
-                        -- name setting but not the variables it interpolates
+                        -- the tab label comes from its active session's name,
+                        -- and the split gets none of its own: it copies the
+                        -- parent's profile but not the session variables that
+                        -- profile's name may interpolate
                         set name of newSession to overlayTitle
                         return id of newSession
                     end if
--- a/.claude-plugin/skills/revdiff/scripts/launch-revdiff.sh
+++ b/.claude-plugin/skills/revdiff/scripts/launch-revdiff.sh
@@ -518,12 +518,13 @@ LAUNCHER
     ITERM_UUID="${ITERM_SESSION_ID##*:}"
 
     # find target session by UUID, auto-detect split direction, capture new session id
-    ITERM_NEW_SESSION=$(osascript - "$ITERM_UUID" "$LAUNCH_SCRIPT" "$CWD" "$SENTINEL" <<'APPLESCRIPT' 2>&1
+    ITERM_NEW_SESSION=$(osascript - "$ITERM_UUID" "$LAUNCH_SCRIPT" "$CWD" "$SENTINEL" "$OVERLAY_TITLE" <<'APPLESCRIPT' 2>&1
 on run argv
     set targetId to item 1 of argv
     set launchScript to item 2 of argv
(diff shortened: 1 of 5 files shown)


umputun:
two gaps against what #307 proposed.

`plugins/revdiff-planning/scripts/launch-plan-review.sh` needs the same edit and doesn't get it here. Its iTerm2 branch at line 385 splits without naming the session, same as the two you fixed, and it computes its own `OVERLAY_TITLE` at line 73. Your issue lists it. Without that edit plan reviews keep the behavior this PR fixes for diff reviews.

no test. The issue proposed an args-capture assertion in the `FAKE_AGTERM_ARGS_FILE` style on the `iterm2` case, and that still looks right. Nothing in the suite would notice if the title argument were dropped again.

on the results table, I couldn't reproduce either the empty pane name or the window title moving. Measurements are in #307. The pane falls back to the profile's default title rather than rendering empty, and the window title never changed here at all. Worth trimming the claim to what the fix actually restores, the tab label.

umputun (approved):
lgtm

